PR #12, opened 2019-09-21
ShellCheck Changes

qubitrenegade:
This is largely the result of changes affected as a result of running [shellcheck](https://github.com/koalaman/shellcheck)

In a nutshell:

* Prefer lower case variable names so as to avoid confilcts with the environment variables
* don't export unless necessary
* allow variables to be overridden 
* make readonly variables readonly (`local readonly`* doesn't mean anything in bash. you need `local -r variablename` to instantiate a readonly variable) ((actually, `local readonly var=10` defines two variables, readonly which in undefined and var which is 10))

I've also added an attempt at a CircleCI job to run shellcheck based on: https://circleci.com/orbs/registry/orb/circleci/shellcheck  Not sure if it'll run on a branch/fork/PR/...

Also, this PR doesn't solve all shellcheck problems but solves I think the majority of important ones.

Thanks!


bin/navi
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
denisidoro: `<repo>/navi` is the entrypoint of the tool, so we can't move it to `<repo>/bin/navi`: it'll break people's symlinks and the homebrew formula. Please keep it at `<repo>/navi`

scripts/install
```diff
@@ -1,13 +1,12 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-export SCRIPT_DIR="$(cd "$(dirname "$0")/.." && pwd)"
+source "$(dirname -- "$0")/vars"
```
denisidoro: I'm worried how `dirname -- "$0"` will play with the binary created from the homebrew formula, which is a symlink. Could you please test that locally before merging the PR?

denisidoro:
Closing without merging.

For anyone reading this: I'd like to conform with shellcheck but:
- I really need to export some variables (to make the preview window work)
- I really like to use all-uppercase variables to differentiate "global" vs "local" variables

PR #35, opened 2019-09-22
Add k3d cheat

djerfy:
Added cheat for `k3d`

### DIFF
--- a/cheats/k3d.cheat
+++ b/cheats/k3d.cheat
@@ -1,0 +1,34 @@
+% k3d
+
+# Check if docker is running
+k3d check-tools
+
+# Start a subshell for a cluster
+k3d shell
+
+# Create a single k3s cluster in docker containers
+k3d create --name <name>
+
+# Create a multi-node k3s cluster in docker containers
+k3d create --name <name> --workers <count>
+
+# Delete cluster
+k3d delete --name <name>
+
+# Stop cluster
+k3d stop --name <name>
+
+# Start a stopped cluster
+k3d start --name <name>
+
+# List all clusters
+k3d list
+
+# Get kubeconfig location for cluster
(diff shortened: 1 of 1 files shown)


cheats/k3d.cheat
```diff
@@ -0,0 +1,34 @@
...
+k3d shell
+
+# Create a single k3s cluster in docker containers
+k3d create --name <name>
+
+# Create a multi-node k3s cluster in docker containers
+k3d create --name <name> --workers <count>
+
+# Delete cluster
+k3d delete --name <name>
```
denisidoro: Isn't it possible to give the user a finite list of names?

If so, you could write something like:
```
k3d delete --name <cluster>
$ cluster: k3d list names 
```

djerfy: No directly :/
```
╰> k3d list
+-------------+------------------------------+---------+---------+
|    NAME     |            IMAGE             | STATUS  | WORKERS |
+-------------+------------------------------+---------+---------+
| k3s-default | docker.io/rancher/k3s:v0.8.1 | running |   0/0   |
|   grafana   | docker.io/rancher/k3s:v0.8.1 | stopped |   0/3   |
| djerfy-k3s  | docker.io/rancher/k3s:v0.8.1 | stopped |   0/3   |
+-------------+------------------------------+---------+---------+
```

denisidoro: what about `k3d list grep -v | awk '{print $2}' | grep -v "NAME" | awk 'NF {print $0}'`?

denisidoro: Even better, we could simply strip the table borders and use advanced selection. Please refer to the docker example in the README :)

denisidoro: This way, the selection will have more details and a header

Issue #44, opened 2019-09-23
Auto-select on single option


denisidoro:
If there's a single option available, fzf shouldn't be prompted. The single value should be auto-selected.

A `--no-autoselect` could be added as well.

PR #53, opened 2019-09-23
Skip fzf finder if the list contains a single item

nikush:
Hey, thanks for making such a great tool! I stumbled across issue #44 and thought I'd chip in.

My implementation uses fzf's `--select-1` [option][option] to have it skip the finder screen and autoselect the only item in the list.

Using the default cheats that come with navi, it can be tested by starting navi with a query like so:
```bash
navi query "'uncordon"
```
Note: the [single quote][syntax] in the query tells fzf to search for the exact string "uncordon" rather than fuzzy searching for it.

Running the above command will jump you directly into the `kubectl uncordon` command.

Removing the single quote will yield multiple matches and jump you into the fzf finder screen.

Let me know if there's anything more you need in this pull request!

[option]: https://github.com/junegunn/fzf/blob/master/src/options.go#L81
[syntax]: https://github.com/junegunn/fzf#search-syntax

### DIFF
--- a/src/arg.sh
+++ b/src/arg.sh
@@ -4,8 +4,10 @@ ARG_REGEX="<[0-9a-zA-Z_]+>"
 ARG_DELIMITER="£"
 
 arg::dict() {
-   local -r fn="$(awk -F'---' '{print $1}')"
-   local -r opts="$(awk -F'---' '{print $2}')"
+   # capture stdin because subsequent attempts to read it fail
+   local -r stdin="$(cat -)"
+   local -r fn="$(echo "$stdin" | awk -F'---' '{print $1}')"
+   local -r opts="$(echo "$stdin" | awk -F'---' '{print $2}')"
 
    dict::new fn "$fn" opts "$opts"
 }
--- a/src/arg.sh
+++ b/src/arg.sh
@@ -5,7 +5,7 @@ ARG_DELIMITER="£"
 
 arg::dict() {
    # capture stdin because subsequent attempts to read it fail
-   local -r stdin="$(cat -)"
+   local -r stdin="$(cat)"
    local -r fn="$(echo "$stdin" | awk -F'---' '{print $1}')"
    local -r opts="$(echo "$stdin" | awk -F'---' '{print $2}')"
 
--- a/src/ui.sh
+++ b/src/ui.sh
@@ -14,6 +14,7 @@ ui::select() {
    local -r query="$(dict::get "$OPTIONS" query)"
    local -r entry_point="$(dict::get "$OPTIONS" entry_point)"
    local -r preview="$(dict::get "$OPTIONS" preview)"
+   local -r autoselect="$(dict::get "$OPTIONS" autoselect)"
 
    local args=()
    args+=("-i")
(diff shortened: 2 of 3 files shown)


src/ui.sh
```diff
@@ -29,6 +30,9 @@ ui::select() {
       args+=("--header")
       args+=("Displaying online results. Please refer to 'navi --help' for details")
    fi
+   if [ "$autoselect" == true ]; then
```
denisidoro:
```sh
if "$autoselect"; then
```

nikush: Updated!

nikush:
> Cool! Thanks!
> One question, though: do you know if it works when there's a header?
> I mean, if there's a header + 1 possible entry, does autoselection work?

Good question. In theory yes:
```bash
echo -e "HEADER\nvalue" | fzf --header-lines 1 --select-1 # prints: value
```

Though it doesn't seem to be recognising the header when I test this in navi. I'll investigate and see if I can get it to work.

src/arg.sh
```diff
@@ -4,8 +4,10 @@ ARG_REGEX="<[0-9a-zA-Z_]+>"
 ARG_DELIMITER="£"
 
 arg::dict() {
-   local -r fn="$(awk -F'---' '{print $1}')"
-   local -r opts="$(awk -F'---' '{print $2}')"
+   # capture stdin because subsequent attempts to read it fail
+   local -r stdin="$(cat -)"
+   local -r fn="$(echo "$stdin" | awk -F'---' '{print $1}')"
+   local -r opts="$(echo "$stdin" | awk -F'---' '{print $2}')"
```
nikush: This line to read in the "opts" was always failing for me. Looks like Awk tries to read stdin but it was coming through as empty.

nikush:
@denisidoro I was experiencing some trouble with the `arg::dict` function trying to create a dictionary out of an argument and its associated fzf options.

My understanding is that we have 2 awk commands trying to read from stdin, but stdin can only be read from once (I can't explain why that is though), so the second awk command was reading an empty value; essentially dropping the fzf options for the argument.

My solution was to capture stdin into a variable and pipe that into the 2 awk commands.

Do you experience the same problem on your end?

src/arg.sh
```diff
@@ -4,8 +4,10 @@ ARG_REGEX="<[0-9a-zA-Z_]+>"
 ARG_DELIMITER="£"
 
 arg::dict() {
-   local -r fn="$(awk -F'---' '{print $1}')"
-   local -r opts="$(awk -F'---' '{print $2}')"
+   # capture stdin because subsequent attempts to read it fail
+   local -r stdin="$(cat -)"
```
denisidoro: Is the dash required?

nikush: It's not required. I've removed it.

nikush:
@denisidoro no I'm not able to merge the PR myself.

nikush:
I've just noticed that I've got the `ui::select` function reading the 'autoselect' option, but the `arg::pick` function doesn't, so an argument that populates a list with 1 item will still prompt the user for a selection.

I can fix this by moving the reading of the 'autoselect' option into the `ui::pick` function, ensuring that it is globally read.

What do you think?

denisidoro:
These are exactly the two functions I'm not happy with: they have too much scope. Feel free to make any changes that you consider that will improve code quality. And thanks again!

Issue #51, opened 2019-09-23
Support install without sudo permissions?


quadebroadwell:
Is it possible to have an install route without the need for sudo?

UchihaSR:
shall i take this issue?

denisidoro:
@UchihaSR feel free! But I still don't know what path we should use in this case

alxbl:
@quadebroadwell if you use oh-my-zsh, checkout the PR that I submitted yesterday, it will let you use navi as a widget without having to do a system-wide install.

Alternatively, you can just `git clone navi` and manually set your `PATH` to include navi's directory in your dotfiles.

```sh
# .zprofile or similar
export PATH="$PATH:/path/to/navi/repo"
```

You can then manually update using `git pull`.

PR #101, opened 2019-10-01
Add support for oh-my-zsh plugin.

alxbl:
Thanks for making this project, it's really cool! I'd been looking for something like this for a while.
I noticed that your shell widget is pretty much stand-alone and I wanted to be able to use navi directly as an [oh-my-zsh][1] plugin, so I added a few things to the widget and moved it to the root of the repository so that the repo can be cloned straight into the `plugins` folder and detected when added in the plugin array.

I think it should be similarly simple with `fish`, but I'm not too familiar with that shell, so I'll leave it to someone else. The `navi.plugin.zsh` file should still be sourcable as a widget for people not using oh-my-zsh, so technically the `shell/` directory could be removed and `navi widget`updated to use the new file.

## What's included
- oh-my-zsh plugin
- install documentation
- no PATH modification required when used with OMZ
- Added bonus: #51 is kind of unofficially addressed by this (assuming one uses OMZ)

## Future Work
- Pre-filter using `$BUFFER` when hitting `Alt+G` to get something similar to `navi query`
- Update `navi widget zsh` to use `navi.plugin.zsh`

Let me know what you think. I'm working on a few cheatsheets, and will open PRs for those as they become ready.

Cheers,
Alex

[1]: https://github.com/robbyrussell/oh-my-zsh


README.md
```diff
@@ -58,6 +58,37 @@ sudo make install
...
 
+### Using oh-my-zsh
+
+Make sure that your oh-my-zsh `ZSH_CUSTOM` directory is configured,
+then clone navi into the plugins directory.
+
+```sh
+export ZSH_CUSTOM='/path/to/.zsh'
+
+DST="$ZSH_CUSTOM/plugins"
```
denisidoro: what is DST?

README.md
```diff
@@ -58,6 +58,37 @@ sudo make install
...
+mkdir -p "$DST" && cd "$DST"
```
denisidoro: remove `&&` and add a new line (so that if the directory exists, it won't abort)

README.md
```diff
@@ -58,6 +58,37 @@ sudo make install
...
+```
+
+To enable navi, simply add it to the oh-my-zsh plugin array:
+
+```
+# Sample ~/.zshrc
+ZSHCFG="$HOME/.zsh"
+ZSH="$ZSHCFG/oh-my-zsh"
+ZSH_CUSTOM="$ZSHCFG"
+plugins=(archlinux docker compleat git github gpg-agent ssh-agent nvm tmux fzf workon navi)
```
denisidoro: `plugins=(docker tmux navi)`, just to make it more succint

denisidoro:
@alxbl, I've created a PR with more changes: https://github.com/denisidoro/navi/pull/103
Could you check that, please?
(I don't know how to add commits to this PR, given that I don't have write access to your fork 😅)

README.md
alxbl: `DST` stands for destination. It's a local variable (not exported) that is used to simplify the `make && cd` command. I can rename it to something else or inline if you'd prefer.

README.md
alxbl: `mkdir -p` succeeds even if the path already exists, so it won't abort. I can split it into two lines if you prefer, though.

README.md
alxbl: Oops, I copied from my `.zshrc` but forgot to trim the output. Good call.

alxbl:
Your changes looks good to me, if you'd like, I can merge them into my PR, or you can merge your PR after this one? I'll push the changes you requested.

README.md
denisidoro: no strong opinions 👍

README.md
denisidoro: could you rename it to `plugins_dir` or something like that?

denisidoro:
I'd rather you merge my changes into your PR so that `master` will always stay up to date

alxbl:
I've merged all of your changes into the branch. The last thing I'd like to discuss before merging is the OMZ upgrade path:

I think it would make more sense as `cd "$ZSH_CONFIG/plugins/navi" && git pull` considering that `sudo make update` will implicitly call `make install`, which results in symlinks in `/usr/bin` and partly defeats the point of having it as a plugin. Let me know what you think.

README.md
alxbl: Sure

alxbl:
Made the fix from the discussion in #103. Should be all set if you want to take a last look before merging.

Cheers!
Alex

PR #113, opened 2019-10-04
Fix read only error

emersonmx:
I get a error in line 6 of src/ui.sh because the array variable is read only.
I write a fast fix to solve the problem :stuck_out_tongue:


denisidoro:
I don't get it. Why was it read-only if it was simply `local` and not `local -r`?

PR #120, opened 2019-10-11
gpg cheat

jmbataller:
CheatSheet for gpg

### DIFF
--- a/cheats/gpg.cheat
+++ b/cheats/gpg.cheat
@@ -10,30 +10,30 @@ gpg --gen-key
 gpg --list-keys
 
 # distribute public key to key server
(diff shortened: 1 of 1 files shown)


cheats/gpg.cheat
```diff
@@ -0,0 +1,39 @@
...
+gpg --version
+
+# gpg generate key
+gpg --gen-key
+
+# list keys
+gpg --list-keys
+
+# distribute public key to key server
+gpg --keyserver hkp://pool.sks-keyservers.net --send-keys <public-key>
```
denisidoro: What is this for? 🤔

jmbataller: replaced the hardcoded public key server for a variable

Issue #37, opened 2019-09-22
Create widgets for bash+zsh+fish


denisidoro:
Hitting `Ctrl-G` for example could start the script. 

When selecting an entry, instead of executing it, it could fill the prompt (similarly to ctrl-r).

This way the history will be kept clean and there will be the possibility to edit the command as you like. 

Edit: this is **done for zsh, bash and zsh** already!
Please refer to README.md.

![Widget](https://user-images.githubusercontent.com/3226564/65788699-03361080-e132-11e9-9e93-b2a4bc405bed.gif)

denisidoro:
Problem: I don't know how to get input from user in a ZSH widget
Workaround: use fzf with `--print-query`

denisidoro:
In case anyone is interested in trying it out (for ZSH):
```sh
navi_home="$(echo "$(brew --prefix)/navi" || echo "/opt/navi")"
source "${navi_home}/shell/widget.zsh"
# hit alt+g
```

tfriedel:
this is cool! I would love to have this for the fish shell.

adworacz:
Looks like this isn't working for navi installed via Linuxbrew:

```
zshrc:source:176: no such file or directory: /home/linuxbrew/.linuxbrew/navi/shell/widget.zsh
```

```
(19-09-26 10:48:13) <0> [~]
u662b1ab6f1175f % which navi
/home/linuxbrew/.linuxbrew/bin/navi

(19-09-26 10:48:15) <0> [~]
u662b1ab6f1175f % brew --prefix
/home/linuxbrew/.linuxbrew

(19-09-26 10:51:02) <0> [~]
u662b1ab6f1175f % brew info navi
denisidoro/tools/navi: stable 0.9.4
An interactive cheatsheet tool for the command-line
https://github.com/denisidoro/navi
/home/linuxbrew/.linuxbrew/Cellar/navi/0.9.4 (43 files, 107.9KB) *
  Built from source on 2019-09-26 at 10:31:23
From: https://github.com/denisidoro/homebrew-tools/blob/master/navi.rb
==> Dependencies
Build: fzf ✔

(19-09-26 10:51:13) <0> [~]
u662b1ab6f1175f %

```

denisidoro:
@adworacz the path in the instructions were wrong. I've just updated it. Could you confirm if it's working now?

adworacz:
[...]

PR #123, opened 2019-10-11
navi fish plugin with ability to edit commandline

tfriedel:
here's my version of the fish plugin. I adapted it from the fzf plugin which allows to go through the command line history (ctrl-r). the fzf_tmux_height  variables don't seem to have an effect. 
there's also an issue where if you cancel navi (ctrl-c), you still have some invisible ^C / ctrl-c character left in the commandline. So your next command will not work unless you press ctrl-c again. This should probably be fixed somehow. 
I had a problem where the line where I call navi would show prompts when run from the shell, but not in the function. I searched around the fish shell issues and running "stty sane" was suggested as a solution.

### DIFF
--- a/navi.plugin.fish
+++ b/navi.plugin.fish
@@ -1,21 +1,17 @@
-function navi-widget -d 'Call navi'
-  set -q NAVI_USE_FZF_ALL_INPUTS; or set -l NAVI_USE_FZF_ALL_INPUTS "true"
-  begin
-    navi --print | while read -l r; set result $result $r; end
-
-    if [ -n "$result" ]
-      echo $result
-
-      # Remove last token from commandline.
-      commandline -t ""
+function navi_key_bindings
+  function navi-widget -d "Show cheat sheet"
+    set -q FZF_TMUX_HEIGHT; or set FZF_TMUX_HEIGHT 40%
+    begin
+      set -lx FZF_DEFAULT_OPTS "--height $FZF_TMUX_HEIGHT $FZF_DEFAULT_OPTS --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m"
+      stty sane
+      env NAVI_USE_FZF_ALL_INPUTS=true navi --print query (commandline)  | perl -pe 'chomp if eof' | read -lz result
+      and commandline -- $result
     end
+    commandline -f repaint
   end
 
-  # commandline -f repaint
-end
-
-bind \cg navi-widget
-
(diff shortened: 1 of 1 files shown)


denisidoro:
Related: #37

navi.plugin.fish
```diff
@@ -1,21 +1,17 @@
...
-  set -q NAVI_USE_FZF_ALL_INPUTS; or set -l NAVI_USE_FZF_ALL_INPUTS "true"
-  begin
-    navi --print | while read -l r; set result $result $r; end
-
-    if [ -n "$result" ]
-      echo $result
-
-      # Remove last token from commandline.
-      commandline -t ""
+function navi_key_bindings
```
denisidoro: Why are you wrapping everything inside this fn?

I tried running `source $NAVI_HOME/navi.plugin.fish` and it didn't seem to work. Removing this line and the last one makes it work.

navi.plugin.fish
```diff
@@ -1,21 +1,17 @@
...
+  function navi-widget -d "Show cheat sheet"
+    set -q FZF_TMUX_HEIGHT; or set FZF_TMUX_HEIGHT 40%
+    begin
+      set -lx FZF_DEFAULT_OPTS "--height $FZF_TMUX_HEIGHT $FZF_DEFAULT_OPTS --tiebreak=index --bind=ctrl-r:toggle-sort $FZF_CTRL_R_OPTS +m"
```
denisidoro: ✂️ (Actually, is this necessary? Why not let the user decide the defaults given his/her `.fishrc`?)

navi.plugin.fish
tfriedel: I adapted the script from the fzf fish script. The bindings are enabled by a line in ~/.config/fish/functions/fish_user_key_bindings.fish

```
function fish_user_key_bindings
  fzf_key_bindings
  navi_key_bindings
end
```

Otherwise they are always enabled. Not a big deal though, you can also delete the file if you want it disabled.

navi.plugin.fish
tfriedel: It is not necessary. I basically left it from the fzf script without thinking deeply if it's necessary. Personally I never used the shortcut to toggle the shorting mode. In the fzf case the idea was probably that you can different options for different applications of fzf. Now the FZF_TMUX_HEIGHT variable doesn't seem to have an effect anyway. I'm also not sure if navi should support this or not.

denisidoro:
Cool! I'll try to merge this on Monday. Thanks!

Issue #115, opened 2019-10-07
Support skim?


jedahan:
I've been using [skim](https://github.com/lotabout/skim) for a while now, and if I change the commands from `fzf` to `sk`, it invokes fine, but interactive searching doesn't work so well. I'm sure there are some differences between `-i` in both tools, but do not know how to move forward.

denisidoro:
I've never used skim, but it should be straightforward to support it: one just needs to create a "bridge" in [ui::fzf](https://github.com/denisidoro/navi/blob/master/src/ui.sh#L3), translating fzf arguments to skim arguments.

If you're willing to start the work, I can finish it 👍

denisidoro:
I've debugged a little bit and it seems that `skim` doesn't play nice with all arguments from `fzf`. 

The command for `preview` is a little bit different, for instance; `--select-1` won't work; and so on. 

That said, the solution to support remains the same, but it's not as trivial as I imagined (we would need to create custom code to simulate `--select-1` for example.

denisidoro:
Closing because I don't plan to focus on this at the moment. Future readers: feel free to reopen this issue or open a PR! 👍

PR #126, opened 2019-10-14
[WIP] Support skim as an alternative to fzf

jedahan:
First draft at #115

### DIFF
--- a/src/ui.sh
+++ b/src/ui.sh
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+ui::fzf_cmd() {
+  [ "$NAVI_ENV" == "test" ] && echo fzf_mock && return
+  command_exists sk && echo sk && return
+  echo "fzf -i"
+}
+
 ui::fzf() {
    local -r autoselect="$(dict::get "$OPTIONS" autoselect)"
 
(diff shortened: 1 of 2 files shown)


src/ui.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+ui::fzf_cmd() {
+  [ "$NAVI_ENV" == "test" ] && echo fzf_mock && return
+  command_exists sk && echo sk && return
+  echo "fzf -i"
```
jedahan: untested, kinda hacky

denisidoro: I'm not sure I got it... You're changing the behavior for `fzf` here? Doesn't it work as-is?

denisidoro:
Related: https://github.com/denisidoro/navi/issues/115

src/ui.sh
jedahan: Actually its the opposite - i'm removing "-i" from skim, but keeping it in `fzf`.

denisidoro:
Closing without merging because I don't plan to support skim in the bash version of navi. Maybe I will in the Rust version I'm planning to write.

Feel free to reopen the PR, though!

PR #138, opened 2019-10-24
Refactor list visualization

### LINKED ISSUE #137: Mixed comments and snippets in search mode
# Issue Description
## Environment
- System: Ubuntu18.04
- Installation method: linuxbrew

When search a snippet, the comments and commands of all candidate snippet are interleved in a random way. 
Please see snapshot below :)

image

denisidoro:
Fixes #137 

Demo

### DIFF
--- a/src/cheat.sh
+++ b/src/cheat.sh
@@ -41,16 +41,18 @@ cheat::memoized_read_all() {
       | cheat::join_lines
 }
 
+# TODO: move this elsewhere
 cheat::get_index() {
-  local -r txt="$1"
-  local -r ref="$2"
+   local -r txt="$1"
+   local -r ref="$2"
 
-  local -r i="$(echo "$txt" | grep "${ref}\$" | awk '{print $1}')"
-  echo $((i - 1))
+   local -r i="$(echo "$txt" | grep "${ref}\$" | awk '{print $1}')"
+   echo $((i - 1))
 }
 
+# TODO: move this elsewhere
 cheat::with_nth() {
-  grep -Eo 'with\-nth +([^ ]+)' | awk '{print $NF}'
+   grep -Eo 'with\-nth +([^ ]+)' | awk '{print $NF}'
 }
 
 cheat::prettify() {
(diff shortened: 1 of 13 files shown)


src/cheat.sh
```diff
@@ -42,14 +42,25 @@ cheat::memoized_read_all() {
 }
 
 cheat::prettify() {
-   awk 'function color(c,s) {
-           printf("\033[%dm%s\033[0m",30+c,s)
+  local -r columns="$(tput cols)"
```
denisidoro: use `ui::width`

src/cheat.sh
```diff
@@ -42,14 +42,25 @@ cheat::memoized_read_all() {
...
+   awk \
+     -v COMMENT_MAX=$((columns * 48 / 100)) \
+     -v SNIPPET_MAX=$((columns * 24 / 100)) \
+     -v SEP="$ESCAPE_CHAR_3" \
+     'function color(c,s,max) {
+           if (max > 0 && length(s) > max) {
+              s2=substr(s, 0, max)
```
denisidoro: mutate `s` itself

src/main.sh
```diff
@@ -50,10 +50,14 @@ handler::main() {
 
 handler::preview() {
    local -r query="$1"
-   local -r selection="$(echo "$query" | selection::dict)"
    local -r cheats="$(cheat::memoized_read_all)"
+   local -r selection="$(echo "$query" | selection::dict "$cheats")"
    local -r cheat="$(cheat::from_selection "$cheats" "$selection")"
-   [ -n "$cheat" ] && selection::cmd_or_comment "$selection" "$cheat" | cmd::unescape
+   [ -n "$cheat" ] && (
+      printf '\033[34m# '; selection::comment "$selection" | cmd::unescape
```
denisidoro: extract to another fn

src/ui.sh
```diff
@@ -43,11 +43,16 @@ ui::select() {
 
    echo "$cheats" \
       | cheat::prettify \
+      | column -t -s $(printf "$ESCAPE_CHAR_3") \
```
denisidoro: add fallback to column

denisidoro:
Stop adding ellipsis to commands

wallace11:
Hi,
It looks like a good step in the right direction. Here's how it looks by default on my end:
image

However, there are a couple of concerns.
First, look what happens when searching for terms in the category names:
image

Second, it may be a good idea to move category name to the front since it'd make it look more organized. Something like: `[category] Description <command>`.

Finally, maybe there should be a standard definition to the min width for terminal windows (check out `ncdu`'s implementation for example).
image

On the same issue, maybe also determining column width by percentage and cut the overflowing text

denisidoro:
Hi, in the version you tried out I had already managed to get the terminal width in bash and I set the width of the commentary column to be 60% of the terminal width :+1:

Regarding the tags in front: I'm not sure about it because most of the displayed content would be the tags, especially if they are very big :(

Being aggressive on cutting the overflowing text leads to a clean UI (such as in the image in the PR description) but it makes it impossible to find snippets if you're searching for terms that have been dropped

The only good solution I can think of is to be able to search by words that are not displayed on the row, but AFAIK fzf doesn't have that option

wallace11:
I understand that there are limits bound by fzf.
The only solution that I can think of atm is having interchangeable `--hide-description` and `--hide-command` options. Afaik, fzf supports setting independent previews so the they should show both regardless of the option selected.

denisidoro:
What if the row were "comment tags snippet" by default or "snippet tags comment" if --snippet-first is set? 

This way one can always avoid the problem you stated (searching for something that is in the far right) 

In addition, there's no information removal (filtering)

wallace11:
I'm not sure that it's going to solve this problem entirely.
Due to the fuzziness nature of fzf's algorithm, words from the far right may match the search term no matter what.

denisidoro:
Fixes #137 

Demo

PR #139, opened 2019-10-26
Additional cheat files

spoki0:
Some of the most common commands for Java Keytool and OpenSSL, added to the cheats.


cheats/keytool.cheat
```diff
@@ -0,0 +1,58 @@
+% java keytool, certificate, encryption
+
+## Creating
```
denisidoro: Cool! I didn't think of "nested" comments yet it actually works :)

PR #155, opened 2019-11-05
Add keybinding to copy snippet

### LINKED ISSUE #152: Hotkey to yank command / snippet to clipboard
i have an alias for `navi --print | xclip -sel clip` but just wanted to suggest adding a hotkey like ctrl-y for this

denisidoro:
Fixes #152

### DIFF
--- a/src/clipboard.sh
+++ b/src/clipboard.sh
@@ -11,5 +11,8 @@ clip::set() {
       echo "$input" | xclip -sel clip
    elif command_exists xsel; then
       echo "$input" | xsel -i -b
+   else
+      echo "$input"
+      die "Unable to set clipboard"
    fi
 }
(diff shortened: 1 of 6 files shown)


src/clipboard.sh
```diff
@@ -0,0 +1,15 @@
...
+
+   if command_exists pbcopy; then
+      echo "$input" | pbcopy
+   elif command_exists clip.exe; then
+      echo "$input" | clip.exe
+   elif command_exists xclip; then
+      echo "$input" | xclip -sel clip
+   elif command_exists xsel; then
+      echo "$input" | xsel -i -b
+   fi
```
zweicoder: maybe echo an error or just print to terminal otherwise

PR #156, opened 2019-11-05
Use 2+ spaces as column separator

### LINKED ISSUE #153: Select columns, counting from the right
Due to some outputs being unreliable or with potential for spaces in the value, is it possible to support selection of columns, counting from the right? 

NAME      | VALUE
------------ | -------------
hello         | world
foo bar | baz

As in this example, providing `--- --column 2` would incorrectly fetch "bar" instead of "baz" in the second row.

denisidoro:
Fixes #153

### DIFF
--- a/src/arg.sh
+++ b/src/arg.sh
@@ -79,7 +79,7 @@ arg::pick() {
    if [ -n "$fn" ]; then
       local suggestions="$(eval "$fn" 2>/dev/null)"
       if [ -n "$suggestions" ]; then
-         echo "$suggestions" | ui::fzf --prompt "$arg: " --header-lines "${headers:-0}" | str::column "${column:-}"
+         echo "$suggestions" | ui::fzf --prompt "$arg: " --header-lines "${headers:-0}" | str::column "${column:-}" "${separator:-}"
       fi
    elif ${NAVI_USE_FZF_ALL_INPUTS:-false}; then
       echo "" | ui::fzf --prompt "$arg: " --print-query --no-select-1 --height 1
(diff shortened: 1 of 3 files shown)


spoki0:
Does this by default use 2 spaces, or does it require `--separator 2` or `--separator '\s\s+'` as parameter? I am not able to get it working either way.

Do I need to do anything in particular to restart the plugin if I switch branches?

denisidoro:
It doesn't require any additional configs.

--separator foo should work, though

You just need to change the branch:

cd "$(navi home)" 
git pull
git checkout column/sep

denisidoro:
It doesn't require any additional configs.

--separator foo should work, though

You just need to change the branch:

cd "$(navi home)" 
git pull
git checkout column/sep

spoki0:
I'm getting an error in a previous step of the command now :(

`awk: not an option: --{print $1}`

denisidoro:
There was a typo 🤦‍♂ 

Could you git pull and try again please?

spoki0:
Typos happen to everyone.
Anyways, that seems to work perfectly by default for my the few command cheats I have written. It doesn't split the row where the value have a space, so it is much better.

Testing the `--separator` out, but I'm not sure it works as expected with whitespace values. I keep getting the `unable to fetch suggestions for 'value'!` when trying to use ex. `--separator ' '`. But splitting on an arbitrary symbol otherwise worked without issue.

denisidoro:
I'll merge it as-is and fix this particular issue in another PR. Thanks!

Issue #168, opened 2019-12-15
fish display 


fundon:
<img width="959" alt="Screen Shot 2019-12-16 at 02 54 18" src="https://user-images.githubusercontent.com/27926/70867337-5be9ac80-1faf-11ea-80c6-061836abd8c1.png">

PR #169, opened 2019-12-16
Fix preview in fish shell

dawidd6:
Closes: #168

Might not be the best solution tho, but it does the trick.

### DIFF
--- a/src/ui.sh
+++ b/src/ui.sh
@@ -20,8 +20,10 @@ ui::fzf() {
 ui::select() {
    local -r cheats="$1"
 
+   [[ "$SHELL" =~ 'fish' ]] || local -r sub='$'
+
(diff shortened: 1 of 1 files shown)


src/ui.sh
```diff
@@ -20,8 +20,10 @@ ui::fzf() {
 ui::select() {
    local -r cheats="$1"
 
+   [[ "$SHELL" =~ 'fish' ]] || local -r sub='$'
```
denisidoro: Maybe ${SHELL:-}?

I'm not sure if there's a case where this variable may be unset.

dawidd6: If it's unset, then it is most probably not fish shell, so don't need to provide a default value since it is double-quoted anyway.

Issue #237, opened 2020-03-12
Keep cheats in a separate repository


ngirard:
How about keeping the cheats in their own repository and separating them from Navi ?

denisidoro:
Thanks for the suggestion! 

Just to understand it better: what cons and pros you see if we split the repos?

OriR:
I see it something like Homebrew taps.

There would either be a convention of repository names (`navicheats-mycheats`) or simply a git repo url that navi can understand.

You would use that git repo to download all of the cheats in there, then "import" them to be used within navi (internally, it can still have custom cheats paths, but it would be dynamic according to the place these cheats repos were to be downloaded).
It would allow you to reuse cheats written by others without contributing them to to main navi repo.

Right now it's hard to share cheats and so it's only useful for the more popular tools because they're in the main repo, or you'd have to write your own cheats for the more niche stuff you use.

denisidoro:
Very nice idea! 

Some things that come to my mind:
1. `navi repo add user/repo` could install cheats to `~/.config/navi/cheats/user/repo` for example.
1. We could have a list of repositories as well for the user to choose from.
navi could get the list from a URL such as https://raw.githubusercontent.com/denisidoro/navi/master/list.txt for example.
1. `navi repo add user/repo/sub/folder` could be supported as well.
1. the default cheatsheets could simply be another repository that navi downloads from in the first run.

WDYT?

denisidoro:
Related: https://github.com/denisidoro/navi/pull/238

OriR:
1. Looks good 👍 The only thing is that I wouldn't want to limit it only to GitHub, that's why I suggested a full git repo url.
2. It would be nice as long as it's not limiting to only these repos, the point is to make it decentralized, right? 😉 
3. That's pretty great! that way you can store your cheats along with your dotfiles and people can still these cheats. My only issue here is that, as a maintainer, it's not super obvious that the path to the cheats needs to be constant, otherwise it might break navi.
4. That would be a great next step, it would be better if it could be downloaded during install time.

Another thing I would like to add is the ability to update cheats, something like `navi repo update` to pull all the latest changes for all cheats or `navi repo update user/repo` to update just the one.

Also, listing & removing remote repo cheats might be useful but I don't think it's mandatory in the first phase.

ngirard:
In addition to what has been said, separate repos would also mean separate, cleaner commit histories, that would be easier to follow and help contributors.

itmecho:
[...]

Issue #233, opened 2020-03-12
Use a standardised config location


itmecho:
**Is your feature request related to a problem? Please describe.**
Not a problem as such, more a nicety. Also, only a suggestion!

**Describe the solution you'd like**
I'm the maintainer of the `navi-bin` AUR package and it would be great if `navi` could look for cheat files in a standard location like `~/.config/navi`. This would make it a lot easier to install as you could have `~/.config/navi/builtin` for the bundled cheats and `~/.config/navi/custom` for custom cheats. This way it's not loading the cheats relative to where the binary is located.

**Describe alternatives you've considered**
Currently I'm using the `NAVI_HOME` variable to point navi to my cheats but there isn't really a nice way to set that when installing from the AUR.

You could also have it store the builtin cheats somewhere like `/usr/share/navi/cheats` and then load any other cheats from `~/.config/navi` if that folder exists.

denisidoro:
I'm totally for this!

But I don't know to implement this. Help needed!

Related comments: [here](https://github.com/denisidoro/navi/issues/223#issuecomment-598144375) and [there](https://github.com/denisidoro/navi/issues/223#issuecomment-598249133).

ngirard:
@denisidoro, you can use the [dirs](https://lib.rs/crates/dirs) crate for this. It provides you with `dirs::config_dir()`.

itmecho:
Thanks @ngirard I didn't know about the `data_dir` function! I've added it to my PR =]

denisidoro:
Could you please review https://github.com/denisidoro/navi/pull/258?

PR #238, opened 2020-03-12
Standard Config Directories

itmecho:
First go at adding #233 (Would be cleaner if #237 was implemented)

This is my first rust PR so I'm very open to constructive criticism!

This is an attempt to get `navi` loading config from "standard" config directories. I've split the bundled cheats out into their own folder at the system level and set the fallback cheat path to the user's `XDG_CONFIG_HOME`. I run Linux so haven't been able to test with OSX. It would probably need either some changes in the code or in the `brew` install process to get it working there.

### DIFF
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -4,9 +4,6 @@ use std::fs::File;
 use std::io::{self, BufRead, BufReader, Lines};
 use std::path::{Path, PathBuf};
 
-pub const BUNDLED_CHEAT_DIR: &'static str = "/usr/share/navi/cheats";
-pub const SHELL_PLUGIN_DIR: &'static str = "/usr/share/navi/shell";
-
 pub fn read_lines<P>(filename: P) -> io::Result<Lines<BufReader<File>>>
 where
     P: AsRef<Path>,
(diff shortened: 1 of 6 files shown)


itmecho:
I've pushed a new commit here to use the `data_dir` function from the `dirs` crate instead of the consts I initially used which is a more flexible approach and doesn't require any elevated permissions to install

src/filesystem.rs
```diff
@@ -34,31 +34,41 @@ fn follow_symlink(pathbuf: PathBuf) -> PathBuf {
...
-    let array = ["cheats", "../libexec/cheats", "../cheats", "../../cheats"];
-    for elem in &array {
-        let p = format!("{}/{}", exe_parent_str, elem);
-        let meta = metadata(&p);
-        if let Ok(m) = meta {
-            if m.is_dir() {
-                return Some(PathBuf::from(p));
-            }
+pub fn bundled_cheat_pathbuf() -> Option<PathBuf> {
+    match dirs::data_dir() {
```
denisidoro: I'm afraid there's no way for the brew formula to save files to `dirs::data_dir()` 😞

itmecho: Ah that's a shame!

I'm not sure if it's considered good practice or not but we could have 2 versions of that function and only compile one for macos and one for not macos using the cfg attributes?

Something like 

```
#[cfg(target_os = "macos")]
pub fn bundled_cheat_pathbuf() -> Option<PathBuf> {
...
}

#[cfg(not(target_os = "macos"))]
pub fn bundled_cheat_pathbuf() -> Option<PathBuf> {
...
}
```

denisidoro:
Closing without merging in favor of #258. Thanks again for the initial work!

PR #248, opened 2020-03-13
Update javascript cheats

toan2406:
Update javascript cheat with `yarn` and run script command

### DIFF
--- a/cheats/javascript.cheat
+++ b/cheats/javascript.cheat
@@ -21,6 +21,38 @@ npm install <package_name> --save-dev
 # install globally a specified package
 npm install <package_name> -g
 
+# run a script
+npm run <script>
+
+$ script: cat package.json | python -c "import sys, json; print('\n'.join(json.load(sys.stdin)['scripts'].keys()))"
+
+
+
+% yarn, node, js
+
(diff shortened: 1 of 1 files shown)


cheats/javascript.cheat
```diff
@@ -21,6 +21,38 @@ npm install <package_name> --save-dev
 # install globally a specified package
 npm install <package_name> -g
 
+# run a script
+npm run <script>
+
+$ script: cat package.json | python -c "import sys, json; print('\n'.join(json.load(sys.stdin)['scripts'].keys()))"
```
OriR:
```
node -p "Object.keys(require('./package.json').scripts).join('\n')"
```

If you're searching for `node` snippets, you'll probably have `node`, not necessarily `python` 😃

toan2406: lol, that makes sense. Let me update it

OriR: Also, you don't need the `cat` now, the `require('./package.json')` part reads the package.json file for you 🙃

toan2406: Thanks. I forgot that

Issue #257, opened 2020-03-13
Update official brew formula


denisidoro:
I started it here: https://github.com/Homebrew/homebrew-core/pull/51359

I want things to become more stable before someone merging it, though.

Probably after https://github.com/denisidoro/navi/issues/237.

denisidoro:
https://github.com/Homebrew/homebrew-core/pull/51772

Issue #256, opened 2020-03-13
Add to crates.io


denisidoro:
The way I see, this is blocked by https://github.com/denisidoro/navi/issues/237 because it's not trivial to bundle the `.cheat`s with the binary

Issue #140, opened 2019-10-26
Hide snippets for unavailable commands


denisidoro:
If yum isn't installed, for example, all snippets that start with yum should be hidden

1) this shouldn't impact startup time
2) if --show-all is used, there should be no filtering

enisozgen:
This is really good idea to achieve since project is grooving well. 

I guess more easy to implement, you can define some kind of `NAVI_IGNORE` variable and users can add their groups there.
 
`NAVI_IGNORE='yum, mac, k8, kubernetes, etc, etc'`

In This case it will more more static than yours but still useful

Best Regards

denisidoro:
A friend of mine suggested to, instead of checking the availability of a command given its snippet,  introduce a new syntax, as:

```
% version control
& git

# Change branch
git checkout <branch>
```

`& git` would determine that this cheat will only appear if `git` is installed, for example

spoki0:
Unless I'm wrong, `which git` can be used to verify that the tool is installed on Linux as well.

OriR:
@denisidoro 
If your leaning towards the new syntax, then it  should be applicable for tag section, command and variables.

let's say you have
```
% git

# print current commit
git show HEAD

# copy current commit
git show HEAD | pbcopy
```

If you were to add this new syntax only on the tag level you'd hide both commands because it'll probably be something like `& git, pbcopy`

If you were to add it only on the lowest level (commands & variables) it would be a huge copy/paste problem, for instance, in the example each command would need git.

I think adding them to all levels and then "merging" them would be the best in terms of UX.
If all the requirements in the tag section pass, you can look at the requirements in the command/variable and if all of them pass, you can show the snippet.

denisidoro:
Could you please review https://github.com/denisidoro/navi/pull/258 as a possible workaround?

denisidoro:
I'm closing this because there are workarounds for the purpose of hiding unwanted snippets

Issue #52, opened 2019-09-23
Deploy to AUR


gunar:
Would love to have this on Arch's User Repository ([AUR](https://aur.archlinux.org/)).

Let me know if you want me to write and maintain a git-based MAKEFILE for you ([example](https://aur.archlinux.org/packages/reredirect-git/)).

denisidoro:
Hi! Unfortunately, I'm not sure if I can commit myself to go down the path of maintaining packages for multiple package managers (besides brew).

That said, I would really appreciate if someone could maintain it! :)

wallace11:
[There you go](https://aur.archlinux.org/packages/navi/), wish granted.

Note to dev:
Current installation method makes it extremely clumsy to package. 
I recommend using standard `make` procedures instead of a custom script. Launcher script could live as a file in the repo - I had to create one from the installation script.

Anyways, thanks for this great useful tool. I hope to contribute some PRs in the near future, as time allows :)

gunar:
Thanks @wallace11!

denisidoro:
For reference, https://aur.archlinux.org/packages/navi-bin/ has been created with a newer version.

Ideally, we should keep https://aur.archlinux.org/packages/navi/ only but I don't know if that's possible.

wallace11:
I was hoping you'd give me enough time to reply on the AUR pkg comments. 
@denisidoro Are you itmecho on AUR? If that's the case, then I can make you a co-maintainer over there. 

Anyways, I'm out of town for about a week and won't be able to do anything in the meanwhile. Please hold on as I'll take care of everything as soon as I get back.

denisidoro:
@wallace11, no problem! Take your time!

No, I'm not itmecho. But I found him/her on Reddit if that's useful: https://www.reddit.com/r/rust/comments/fgzifj/-/fkasq7d

itmecho:
Hi @wallace11 ! I'm itmecho everywhere =] 

I created `navi-bin` to provide the precompiled binary as an option for people who just want the binary without the dependency on rust and the compile time. The `navi` package should remain as the compiled from source version. This seems to be a common pattern, for example [starship's](https://github.com/starship/starship) AUR packages. Would you agree that it's fine to keep both of these AUR packages to give people the option?

I'm happy to make you a co-maintainer on the `navi-bin` package as well if you want!

As @denisidoro said though, I'm in no rush so take your time =]

itmecho:
[...]

PR #258, opened 2020-03-14
Initial cheat repo support

### LINKED ISSUE #233: Use a standardised config location

denisidoro:
This PR makes navi stop bundling `.cheat` files and makes it able to download from repos on GitHub, using a standardized config directory.

- it fixes #233 and #237
- it makes #52, #256 and #257 much easier to implement
- it makes #40 harder to implement
- it's an alternate solution to #140 
- it's influenced by #238 and #254.

This PR ended up being much bigger than I would like so if you could review only this description it would be already very nice!

When navi is started for the first time, a welcome cheatsheet is shown:
Screenshot at 2020-03-15 10-20-04

If the user selects the first option, the user is prompted to download `.cheat`s from https://github.com/denisidoro/cheats: 
Screenshot at 2020-03-15 10-20-35
Screenshot at 2020-03-15 10-22-59

The config folder is populated:
Screenshot at 2020-03-15 10-21-11

When run again, cheats are available:
Screenshot at 2020-03-15 10-21-34

### Breaking changes

In order to make navi stop bundling shell widgets as well, a breaking change has been introduced: `source $(navi widget bash)` won't work anymore. It should be `source <(navi widget bash)` now. Any ideas on how to make this transition more graceful?


src/cmds/repo.rs
```diff
@@ -0,0 +1,73 @@
...
+    let repo_folder_str = format!("{}-master", repo);
+    let tmp_path_str = format!("{}/tmp", cheat_path_str);
+
+    // TODO: stop shell'ing out
+    let cmd = format!(
+        r#"
+      url="https://github.com/{user}/{repo}/archive/master.tar.gz";
+      mkdir -p "{tmp_path}"; 
+      cd "{tmp_path}"; 
+      (wget -c "$url" -O - || curl -L "$url") | tar -xz || exit 42;
```
denisidoro: Is this compatible with most systems? :thinking:

OriR: Any reason not to use [`git clone --depth 1`](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt)?

`git` command will probably be available in all platforms, plus you don't need to extract the gz file later, yet you don't get all of the git history, just the latest commit.
In the future, it'll be easier to get the latest changes with `git pull --depth/--deepen` without having to download each repo all over again.

src/cmds/shell.rs
```diff
@@ -1,19 +1,13 @@
...
-
 pub fn main(shell: &str) -> Result<(), Box<dyn Error>> {
-    let file = match shell {
-        "zsh" => "navi.plugin.zsh",
-        "fish" => "navi.plugin.fish",
-        _ => "navi.plugin.bash",
+    let content = match shell {
+        "zsh" => include_str!("../../shell/navi.plugin.zsh"),
+        "fish" => include_str!("../../shell/navi.plugin.fish"),
+        _ => include_str!("../../shell/navi.plugin.bash"),
```
OriR: Isn't there a way to create a filesystem path that's platform specific?
Windows would prefer `\` rather than `/` (even though both work).

src/cmds/repo.rs
denisidoro: Because I didn't want to add dependencies for a curl-like and tar-like operations. I may reconsider it though

README.md
```diff
@@ -91,13 +91,13 @@ You can use **navi** as a widget to your shell. This way, your history is correc
 In order to use it, add this line to your `.bashrc`-like file:
 ```sh
 # bash
-source "$(navi widget bash)"
+source <(navi widget bash)
```
OriR: I'm not super versed in bash/zsh/fish/rust, but I couldn't find the reason in the code that required this change.

AFAIK `source` needs a path to, well, source in the current shell.
This, [if I understand it correctly](http://www.gnu.org/software/bash/manual/bash.html#Process-Substitution), reads from the given file and passes it as the stdin for `source`, which might work but I think is unnecessary.

src/cmds/repo.rs
```diff
@@ -0,0 +1,74 @@
...
+    let user = parts.next().unwrap();
+    let repo = parts.next().unwrap();
+    let cheat_path_str = filesystem::pathbuf_to_string(filesystem::cheat_pathbuf().unwrap());
+    let repo_folder_str = format!("{}-master", repo);
+    let tmp_path_str = format!("{}/tmp", cheat_path_str);
+
+    // TODO: stop shell'ing out
+    let cmd = format!(
+        r#"
+      url="https://github.com/{user}/{repo}/archive/master.tar.gz";
```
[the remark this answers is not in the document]
denisidoro: I'll add support for adding arbitrary urls before merging this, then :+1:

src/cmds/repo.rs
Kibouo: This way there is a silent dependency on the above commands, assuming they're present. Some systems, *and users*, like to do weird stuff :man_shrugging:

denisidoro: Regarding git clone: fewer systems have git (compared to curl/wget) + the repo source code is temporary the way I implemented, so a git pull wouldn't hep

README.md
denisidoro: Previously the command returned a path to a shell script. Now it prints the content of the shell script.

src/cmds/shell.rs
denisidoro: This is at compile time. It's the same as if I copy pasted the contents of file to the source code. I think rust handles the / vs \ thing

OriR: Ohhh, didn't know that `include_str` macro actually reads the file in _compile time_!

So, now that we're passed that, why was this done?
I'm assuming that's because you didn't want to add these files as external resources so it'll be easier to bundle with additional packaging methods?

If that's the case, then you can write that file to a similar place like the cheats and then print the path to that file.

But, it's a super simple change from the users perspective so I wouldn't bother going to all that trouble.
Plus, it's easier to change the bound key if you want to - instead of reading the output of the command and then regexing & replacing the key, you just need to regex & replace the key.

PR #259, opened 2020-03-14
add `--allow-extra` option

### LINKED ISSUE #253: $ variables as "recomendation"

in this snippet

```
# login to a server and forward to ssh key (dangerous but usefull for bastion hosts)
ssh -A <user>@<server>
```

I would like to give just "recommendations" for `user`.

This just sets it to `whoami`
```
$ user : whoami 
```

This opens fzf and let me type something but typing something like `root` lets `navi` crash
```
$ user : whoami && echo 
```
```
thread 'main' panicked at 'byte index 18446744073709551615 is out of bounds of ``', src/libcore/str/mod.rs:2017:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

I also tried 
```
$ user : whoami && echo --- --print-query
```
but this resulted in the same error.

I would like to write this
```
$ user : whoami && echo "root" --- --print-query
```
or
```
$ user : whoami --- --print-query --no-select-1
```

to use `whoami` and `root` only as recommendation, so I can type what ever I want.

mrVanDalo:
Closes #253.

I had to make some refactoring, to write some tests on the way to understand the flow. It ended up with more changes, than I wanted.
I also ran `cargo fmt`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 /target
 **/*.rs.bk
+
+# intellij
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,2 @@
 /target
 **/*.rs.bk
-
-# intellij
-.idea
--- a/README.md
+++ b/README.md
@@ -176,7 +176,13 @@ $ y: echo -e "$((x+10))\n$((x+20))"
 
 ### Variable options
 
-For lines starting with `$` you can add extra options using `---`.
+For lines starting with `$` you can add use`---` to parse parameters to `fzf`.
+* `--allow-extra` : handles `fzf` option `--print-query`. `enter` will prefer a selection,
+    `tab` will prefer the query typed.
+* `--multi` : forwarded option to `fzf`.
+* `--header` | `--headers` | `--header-lines` : forwarded option to `fzf`
+* `--column` : forwarded option to `fzf`.
+* `--delimiter` : forwarded option to `fzf`.
 
 #### Table formatting
 
--- a/README.md
+++ b/README.md
@@ -177,10 +177,10 @@ $ y: echo -e "$((x+10))\n$((x+20))"
 ### Variable options
 
 For lines starting with `$` you can add use`---` to parse parameters to `fzf`.
-* `--allow-extra` : handles `fzf` option `--print-query`. `enter` will prefer a selection,
-    `tab` will prefer the query typed.
+* `--allow-extra` *(experimental)*: handles `fzf` option `--print-query`. `enter` will prefer a selection,
+    `tab` will prefer the query typed. 
 * `--multi` : forwarded option to `fzf`.
-* `--header` | `--headers` | `--header-lines` : forwarded option to `fzf`
+* `--header-lines` : forwarded option to `fzf`
 * `--column` : forwarded option to `fzf`.
 * `--delimiter` : forwarded option to `fzf`.
 
--- a/src/cheat.rs
+++ b/src/cheat.rs
@@ -12,10 +12,10 @@ pub struct SuggestionOpts {
     pub header_lines: u8,
     pub column: Option<u8>,
     pub delimiter: Option<String>,
-    pub suggestion_type : SuggestionType,
+    pub suggestion_type: SuggestionType,
 }
 
-#[derive(Clone,Copy,Debug,PartialEq)]
+#[derive(Clone, Copy, Debug, PartialEq)]
 pub enum SuggestionType {
     Disabled,
     SingleSelection,
@@ -65,13 +65,12 @@ fn parse_opts(text: &str) -> SuggestionOpts {
         column,
         delimiter,
         suggestion_type: match (multi, allow_extra) {
-            (true,_) =>  SuggestionType::MultipleSelections , // multi wins over allow-extra
-            (false,true) => SuggestionType::SingleRecommendation,
-            (false,false) => SuggestionType::SingleSelection
+            (true, _) => SuggestionType::MultipleSelections, // multi wins over allow-extra
+            (false, true) => SuggestionType::SingleRecommendation,
+            (false, false) => SuggestionType::SingleSelection,
         },
     };
     result
-
 }
 
 fn parse_variable_line(line: &str) -> (&str, &str, Option<SuggestionOpts>) {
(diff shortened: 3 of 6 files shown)


mrVanDalo:
@denisidoro I implemented the `tab` `enter` behavior a bit different. Because the user of the template is not capable to see if `--allow-extra` is set or not, so the user would expect `enter` works the same. so typing `peter` `enter` will just return `peter`.

I'm using this setup for quite some time, and I think the `--allow-extra` setup be used all the time, but this is another discussion.

.gitignore
```diff
@@ -1,2 +1,5 @@
 /target
 **/*.rs.bk
+
+# intellij
```
denisidoro: I'm fine with this but I recommend you to setup a global .gitignore in your system

README.md
```diff
@@ -176,7 +176,13 @@ $ y: echo -e "$((x+10))\n$((x+20))"
 
 ### Variable options
 
-For lines starting with `$` you can add extra options using `---`.
+For lines starting with `$` you can add use`---` to parse parameters to `fzf`.
+* `--allow-extra` : handles `fzf` option `--print-query`. `enter` will prefer a selection,
+    `tab` will prefer the query typed.
+* `--multi` : forwarded option to `fzf`.
+* `--header` | `--headers` | `--header-lines` : forwarded option to `fzf`
```
denisidoro: Header and headers are for backwards compatibility only. Theyre not intended to be used

README.md
```diff
@@ -176,7 +176,13 @@ $ y: echo -e "$((x+10))\n$((x+20))"
 
 ### Variable options
 
-For lines starting with `$` you can add extra options using `---`.
+For lines starting with `$` you can add use`---` to parse parameters to `fzf`.
+* `--allow-extra` : handles `fzf` option `--print-query`. `enter` will prefer a selection,
```
denisidoro: Please add [alpha] or [experimental] because I'm still thinking about this API. maybe itll be the default in the future

src/cheat.rs
```diff
@@ -161,3 +183,49 @@ pub fn read_all(config: &Config, stdin: &mut std::process::ChildStdin) -> HashMa
 
     variables
 }
+
+#[cfg(test)]
```
denisidoro: Could you please move the tests to the /tests folder?

src/cheat.rs
```diff
@@ -61,17 +82,15 @@ fn parse_variable_line(line: &str) -> (&str, &str, Option<SuggestionOpts>) {
     let caps = re.captures(line).unwrap();
     let variable = caps.get(1).unwrap().as_str().trim();
     let mut command_plus_opts = caps.get(2).unwrap().as_str().split("---");
-    let command = command_plus_opts.next().unwrap();
-    let opts = match command_plus_opts.next() {
-        Some(o) => Some(parse_opts(o)),
-        None => None,
-    };
-    (variable, command, opts)
+    let command: &str = command_plus_opts.next().unwrap();
```
denisidoro: Is the type necessary?

src/cheat.rs
```diff
@@ -61,17 +82,15 @@ fn parse_variable_line(line: &str) -> (&str, &str, Option<SuggestionOpts>) {
...
+    let command_option_string: Option<&str> = command_plus_opts.next();
```
denisidoro: Maybe use this value directly instead of using let?

src/cheat.rs
```diff
@@ -161,3 +183,49 @@ pub fn read_all(config: &Config, stdin: &mut std::process::ChildStdin) -> HashMa
...
+        let path = "tests/cheats/ssh.cheat";
+        let mut variables: HashMap<String, Suggestion> = HashMap::new();
+        let mut child = Command::new("cat").stdin(Stdio::piped()).spawn().unwrap();
+        let child_stdin = child.stdin.as_mut().unwrap();
+        read_file(path, &mut variables, child_stdin);
+        let mut result: HashMap<String, (String, std::option::Option<_>)> = HashMap::new();
+        result.insert(
+            "ssh;user".to_string(),
+            (
+                " echo -e \"$(whoami)\\nroot\" ".to_string(),
```
denisidoro: Maybe use raw strings to avoid escaping?

src/fzf.rs
```diff
@@ -132,7 +134,87 @@ where
...
+                (Some(one), Some(termination), None) if *termination == "enter" => one.to_string(),
+                (Some(one), Some(termination), _) if *termination == "tab" => one.to_string(),
+                _ => "".to_string(),
+            }
+        }
+    }
+}
 
-    (text, result)
+#[cfg(test)]
```
denisidoro: Please move tests

src/cmds/core.rs
```diff
@@ -171,11 +176,11 @@ fn replace_variables_from_snippet(
 pub fn main(variant: Variant, config: Config, contains_key: bool) -> Result<(), Box<dyn Error>> {
     let _ = display::WIDTHS;
 
-    let (raw_output, variables) = fzf::call(gen_core_fzf_opts(variant, &config), |stdin| {
+    let (raw_snippet, variables) = fzf::call(gen_core_fzf_opts(variant, &config), |stdin| {
```
denisidoro: I call "snippet" the bash command. This would be like raw_selection or raw_line

README.md
mrVanDalo: Ok so only `--header-lines` stays.

src/cheat.rs
mrVanDalo: sure, can do, but being able of putting the tests in the same file then the code, is one of the best features of rust.

src/cheat.rs
mrVanDalo: not really, can remove it.

src/cheat.rs
mrVanDalo: can do.

mrVanDalo:
Alright, will do your comments.

src/cheat.rs
mrVanDalo: I wanted to point out, I have to make functions public to make the tests work, which don't need to be public.

mrVanDalo: https://doc.rust-lang.org/stable/book/ch11-03-test-organization.html#testing-private-functions
I follow these recommendation/descriptions.

mrVanDalo: Please let me know if I really should move the tests to `/tests` and make all the functions tested public.

denisidoro: Wow! Didn't know this was idiomatic in Rust. It seems very counter-intuitive from my experience hahaha

But there must be good reasons for that (I haven't read the documents yet, sorry) so ignore my previous comment :+1:

mrVanDalo: :+1: 
That was actually one of the first features that draw my attention to rust, because tests are one of the best ways to document code.

mrVanDalo:
@denisidoro yeah also testing a bit would be not wrong. I tested quite intense, but still there could have slipped something through my attention. I'm not that fit with `fzf` the `--multi` selection and `--header-lines` stuff and all.

PR #274, opened 2020-03-16
fix openssl dependency and introduce travis

mrVanDalo:
The openssl dependency is unnecessary. I created a feature dependency for macos targets.

I also created a travis ci build script to build the project on linux and macos machines
to verify if everything is still working.

@denisidoro to make the travis part work, you have to sign in to travis-ci.org (can do that via github) and enable this project. Everything is already linked properly on the README and crates.io.

Have a look at [my travis builds](https://travis-ci.org/github/mrVanDalo/navi) you can see linux and osx build just fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,54 @@
+# https://travis-ci.org/denisidoro/navi
+language: rust
+
+jobs:
+  include:
+
+  - name: run tests
+    rust: stable
+    script:
+    - cargo test --all --locked
+    - rustup component add rustfmt-preview
+    - cargo fmt --all -- --check
+    env: RUST_BACKTRACE=1
+
+  - name: build binary on ubuntu
+    rust: stable
+    script: cargo build --release --locked
+
+  - name: build binary for x86_64-unknown-linux-musl
+    rust: stable
+    script:
+      - cargo install cross
+      - cross build --release --locked --target x86_64-unknown-linux-musl
+
+  - name: build binary for aarch64-linux-android
+    rust: stable
+    script:
+      - cargo install cross
+      - cross build --release --locked --target aarch64-linux-android
+
+  - name: build binary for armv7-linux-androideabi
+    rust: stable
+    script:
+      - cargo install cross
+      - cross build --release --locked --target armv7-linux-androideabi
+
+  - name: build binary for armv7-unknown-linux-musleabihf
+    rust: stable
+    script:
+      - cargo install cross
+      - cross build --release --locked --target armv7-unknown-linux-musleabihf
+
+  - name: build binary on osx
+    os: osx
+    rust: stable
+    script: cargo build --release --locked
+
+notifications:
+  irc:
+    on_failure: always
+    channels:
+      - "irc.freenode.net#navi"
+    template:
+      - "%{result} - %{repository_slug} %{branch} \"%{commit_subject}\" %{build_url}"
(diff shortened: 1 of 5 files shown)


denisidoro:
Openssl is necessary for the git2 dependency when building using cross: https://github.com/denisidoro/navi/runs/509393805?check_suite_focus=true

mrVanDalo:
Yes, in some cases it is necessary that the `git2` uses `vendored-openssl`, but instead of adding openssl to the dependency list, it is better to configure the `git2` dependency.

As I have done it in this pull request. Like here
```
[target.'cfg(target_os = "macos")'.dependencies.git2]
￼features = [ "vendored-openssl" ]
```
I guess for the `android` `target_os` it needs theses features also enabled (the link you posted is about a cross compiled `android` build)

mrVanDalo:
@denisidoro I will add another travis job to test if cross compiling works for aarch64 as well. And change Cargo.toml so this part is covered as well. Give me 30 min

denisidoro:
Could you please make sure it works with ARM and Android, if possible? 

I've closed https://github.com/denisidoro/navi/pull/277.

mrVanDalo:
https://travis-ci.org/github/mrVanDalo/navi/jobs/663033712 < @denisidoro looks like aarch64-linux-android builds without problems. (just one warning)

mrVanDalo:
Ah I see you cross compile also the linux x86_64 version. Why is that? isn't the machine which builds the stuff exactly that architecture?

denisidoro:
Because I compile with the musl target. If I use the Ubuntu image and compile it without cross, the resulting image won't work on alpine-linux, for example.

mrVanDalo:
> Because I compile with the musl target. If I use the Ubuntu image and compile it without cross, the resulting image won't work on alpine-linux, for example.

I see, I will change the travis ci builds than as well to musl. I hope the `alpine-linux` thing is not a real use case :D

denisidoro:
Now that you mentioned, let's stop supporting alpine Linux (and not using cross for x86_64 Linux).if anyone complains in the future, we can add another binary. WDYT?

denisidoro:
One question: is it possible to upload these binaries to github releases? If not, I'll still have to use github actions, right?

mrVanDalo:
> Now that you mentioned, let's stop supporting alpine Linux (and not using cross for x86_64 Linux).if anyone complains in the future, we can add another binary. WDYT?

Sounds legit, but i just added a `linux musl` task, I did not replace the `x86_64` build, so no problem here.

mrVanDalo:
> One question: is it possible to upload these binaries to github releases? If not, I'll still have to use github actions, right?

I have no experiences with the github-releases and CI setup, like in general. But I think it should be fairly easy to do so : https://docs.travis-ci.com/user/deployment/releases/
You have to enter a github api key and all, but these can be encrypted, and will only work for `denisidoro/navi` (but not for `mrVanDalo/navi`).

I'm pretty sure, you can also bow github ci to do similar things than travis. I just use travis because I'm it is good documented, and I'm know it quite well.

mrVanDalo:
https://travis-ci.org/github/mrVanDalo/navi < looks all the cross compiling works. About the release process I guess it make sense to create another pull request.

mrVanDalo:
@denisidoro I also add a travis build banner to the README which points to a repository, which exists only when you create a repo at travis. This should be super easy (you can signup with your github account via oauth). But if you for some reason don't want to do that, I can remove the banners I created.

denisidoro:
https://travis-ci.org/github/denisidoro/navi 👍

PR #277, opened 2020-03-16
Use http for git

denisidoro:


mrVanDalo:
I don't think it is a good idea to prefer http over https

denisidoro:
OpenSSL still is required for https 😞 

Any suggestions?

PR #281, opened 2020-03-17
Make --allow-extra default

denisidoro:
Related: https://github.com/denisidoro/navi/pull/259


denisidoro:
@mrVanDalo

mrVanDalo:
Sorry I'm I will not have time before the weekend, because I'm have a flight back home and thinking of maybe stay away from Europe for a bit longer.

PR #287, opened 2020-03-18
Uses a hashset to de-duplicate results

### LINKED ISSUE #283: Deduplicate results
(disclaimer: Hopefully you don't mind all the issues I've been filing. I really enjoy this project so I'm trying to list the things I'm hitting/wishing for :smile: )

This could be a bug or feature-request depending on how you look at it.

`navi` duplicates results. I.e. I have a personal repository of `.cheat` files, as well as a shared repository at work of `.cheat` files. There are .cheat files I have in my personal repo, that are also applicable to my work/co-workers, however I don't want to "move" my personal cheat files into the work repository since I may not be at that job forever, and co-workers can also make changes that my not fit my personal flows.

However, having the same cheats in two places means I get duplicate results cluttering the navigation window and taking time to figure out if they really are duplicates, or just similar (we have a lot of similar cheats as well, that look like duplicates at first glance).

Results are filtered. Exact duplicates are only displayed once.

Only alternatives are manually curating cheat files, or living with duplication.

**Additional context**

Implementation wise, I'd prefer each cheat line to be looked at independently, and not cheat files as a whole. I.e. my personal `git.cheat` isn't an exact duplicate of my work's `git.cheat`, however many cheats inside that file *are*.

I would imagine hashing the the lines would solve this and would be fast enough that it wouldn't impact performance. I'm sure there's some fashion in which the cheats could be lazily hashed as well, (since for example there may be 10,000 cheats, but only 100 or so can fit on a page at once).

I'd be happy to take a look at the impl for this given a starting point of where to look.

kbknapp:
This creates a `HashSet<u64>` using [FNV Hashing](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function), because it's much
faster than Rust's default SIP hash and this isn't used for anything
that needs to be overly cyptographically secure. It also allows us to keep a
hashset of 64bit numbers, instead of Strings or string slices thus
reducing memory consumption.

In unscientific benchmarks of ~5,200 `.cheat` lines containing about
`2,600` duplicate lines there was no noticable increase in memory
consumption (benchmarking time is harder since you wait for human
input), both master (HEAD b78903a9d4fb) and this commit used 10.8mb max
RSS.

To attempt to measure time (and again max RSS) running master against
this commit using the command `navi --print --path dups_folder/ best "git tag"`
resulted in a *decrease* of memory consumption by about 200k, and no
noticeable time increase.

Closes #283

Note, it's also possible to use a `HashSet<String>` if you wish to remove `fnv.rs`, for slightly increased memory and runtime.

### DIFF
--- a/src/cheat.rs
+++ b/src/cheat.rs
@@ -1,9 +1,10 @@
 use crate::display;
 use crate::filesystem;
+use crate::fnv::HashLine;
 use crate::option::Config;
 use crate::welcome;
 use regex::Regex;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::fs;
 use std::io::Write;
 
@@ -111,6 +112,7 @@ fn read_file(
     path: &str,
     variables: &mut HashMap<String, Suggestion>,
     stdin: &mut std::process::ChildStdin,
+    set: &mut HashSet<u64>,
 ) -> bool {
     let mut tags = String::from("");
     let mut comment = String::from("");
@@ -121,6 +123,11 @@ fn read_file(
     if let Ok(lines) = filesystem::read_lines(path) {
         for l in lines {
             let line = l.unwrap();
+            let hash = line.hash_line();
+            if set.contains(&hash) {
+                continue;
+            }
+            set.insert(hash);
 
             // blank
             if line.is_empty() {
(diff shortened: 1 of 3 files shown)


kbknapp:
I should note, I think *technically* we're double hashing (FNV hash of String to u64, then SIP hashing u64), however SIP hashing a u64 is very fast. The only way to get around that is to use something like a Vec<u64> to hold the hash results. That will actually probably run faster than the current FNV/SIP implementation for low numbers of cheats lines (like 5,000), however as the number of cheats grows I'm pretty sure the Hashset will outperform a Vec pretty easily.

kbknapp:
Final note :stuck_out_tongue_winking_eye: It's also possible to use `FnvHasher` as the `HashSet` hasher (instead of SIP), however you still end up with a `HashSet<String>` which means cloning all the lines. Just wanted to mention that since my previous comments gloss over that fact but don't explain why.

denisidoro:
Excellent PR description! :clap:

Regarding the double hashing, we could pass a custom hasher to the HashSet, where hasher = identity.

We can do that in a future PR.

Issue #291, opened 2020-03-19
Improve error reporting


Kibouo:
There is a lot of `unwrap()` used all over the code. It's very hard to locate the reason for crashes when they do happen.

This is related to [this TODO item](https://github.com/denisidoro/navi/projects/1#card-33715856), and I'd like to tackle it.

Questions:
1. To what extent may function signatures be changed? As in making the return type a `Result`? Some `unwrap`s feel uncomfortable and should be bubbled up imho. (I've limited experience working on other people's projects and API changes.)
2. What is your opinion on adding the [anyhow](https://github.com/dtolnay/anyhow) crate? It makes adding context (cause of error) easy, pretty formats error chains, and allows different library error types to interoperate nicely.

denisidoro:
1. I agree the code is very bad and non-idiomatic as-is. This was my first project in Rust and I was learning the language while I was coding it. That said, feel free to make any changes you feel like, including signature changes. I'll happily accept them! :+1: (but please start from the `multi-col` branch in case it hasn't been merged yet)

2. :+1: from me

denisidoro:
Just let me know if you plan on working on this so that our efforts won't overlap, please

Kibouo:
Yes, I will work on it this weekend.On 19 Mar 2020 22:46, Denis Isidoro <notifications@github.com> wrote:
Just let me know if you plan on working on this so that our efforts won't overlap, please

Kibouo:
Closed by #299 🎉

PR #299, opened 2020-03-21
Improve error handling

Kibouo:
PR for #291

- Currently adding error messages based on context. However, I'm not familiar with the code base so **thorough review is highly recommended** to improve error messages
- ... especially for the `fzf` and `core` modules.

TODO:
- [x] Every call to `bash` now has an error message similar to "Failed to execute bash". This should probably be related to the actual command ran, i.e. params given to `bash`.
- [x] A lot of `anyhow::[with_]context()` is used to add, well, context. It might be better to create custom error variants for repetitive errors.
- [x] `navi fn welcome` is broken.
- [x] Add "file a bug report" message to top-level error.
- [x] Test for potential breakages. Logic changes have been avoided as much as possible but stuff happens.
- [x] `navi repo browse` and `navi repo add https://github.com/denisidoro/navi` crash. Root cause is `fzf` not returning data due to window layout error.
image
- [x] bump version

### DIFF
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -9,17 +9,14 @@ use std::fs::File;
 use std::io::{self, BufRead};
 use std::path::{Path, PathBuf};
 
-pub fn read_lines<P>(filename: P) -> Result<Vec<String>, Error>
+pub fn read_lines<P>(filename: P) -> Result<impl Iterator<Item = Result<String, Error>>, Error>
 where
-    P: AsRef<Path> + Display,
+    P: AsRef<Path> + Display + Copy,
 {
-    let error_string = format!("Failed to read lines from `{}`", filename);
-    let file = File::open(filename)?;
-    io::BufReader::new(file)
+    let file = File::open(filename).with_context(|| format!("Failed to open file {}", filename))?;
+    Ok(io::BufReader::new(file)
         .lines()
-        .map(|line| line.map_err(Error::from))
-        .collect::<Result<Vec<String>, Error>>()
-        .with_context(|| error_string)
+        .map(|line| line.map_err(Error::from)))
 }
 
 pub fn pathbuf_to_string(pathbuf: PathBuf) -> Result<String, Error> {
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -1,4 +1,5 @@
 use crate::structures::error::filesystem::InvalidPath;
+use crate::structures::error::filesystem::UnreadableDir;
 use crate::structures::option::Config;
 use anyhow::Context;
 use anyhow::Error;
@@ -78,13 +79,13 @@ fn cheat_paths_from_config_dir() -> Result<String, Error> {
         .and_then(pathbuf_to_string)
         .and_then(|path| {
             fs::read_dir(path.clone())
-                .with_context(|| format!("Unable to read directory `{}`", &path))
+                .map_err(|e| UnreadableDir::new(path.clone(), e).into())
                 .map(|entries| (path, entries))
         })
         .and_then(|(path, dir_entries)| {
             let mut paths_str = String::from("");
             for entry in dir_entries {
-                let path = entry.with_context(|| format!("Unable to read directory `{}`", path))?;
+                let path = entry.map_err(|e| UnreadableDir::new(path.clone(), e))?;
                 paths_str.push_str(
                     path.path()
                         .into_os_string()
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -1,4 +1,5 @@
 use crate::structures::option::Config;
+use anyhow::Error;
 use std::fs;
 use std::fs::File;
 use std::io::{self, BufRead, BufReader, Lines};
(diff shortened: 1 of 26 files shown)


Kibouo:
My approach currently is to:
- eliminate all `unwrap()`, except in tests
- `expect()` when calling an internal function with highly controlled input, since these "ugly" panic messages ~will~ should only ever be seen by devs.
- return as much `Result` as possible, as the added context to each layer of errors will be nicely formatted and hopefully useful for the end-user.

denisidoro:
The `navi repo` bug is fixed in master 👍

Kibouo:
Kindly re-review changes. Please check whether the actual messages are clear.

Also added a wrapping error telling the end-user to file an issue. Ex:
image

This required the error type `FileAnIssue` to be made public in the library. It had to be used in the bin. `handle_config` is called multiple times so it would look weird to potentially have multiple of these messages nested. Is this change ok?

A few things remain to be done as can be seen in the TODO of the first post.

Kibouo:
Would this be a bump in version to 2.4.0?

src/flows/repo.rs
```diff
@@ -2,25 +2,28 @@ use crate::filesystem;
...
-    let repo_path_str = format!("{}/featured", filesystem::tmp_path_str());
+pub fn browse() -> Result<(), Error> {
+    let repo_path_str = format!("{}/featured", filesystem::tmp_path_str()?);
 
-    filesystem::remove_dir(&repo_path_str);
-    filesystem::create_dir(&repo_path_str);
+    // The dir might not exist which would throw an error. But here we don't care about that
+    // since our purpose is to make space for a new directory. Potential other errors
+    // (e.g due to lack of permission) will cause an error during dir creation.
+    let _ = filesystem::remove_dir(&repo_path_str);
```
denisidoro: Why using `let _ =`?

src/flows/repo.rs
```diff
@@ -2,25 +2,28 @@ use crate::filesystem;
...
 use walkdir::WalkDir;
 
-pub fn browse() -> Result<(), Box<dyn Error>> {
-    let repo_path_str = format!("{}/featured", filesystem::tmp_path_str());
+pub fn browse() -> Result<(), Error> {
+    let repo_path_str = format!("{}/featured", filesystem::tmp_path_str()?);
 
-    filesystem::remove_dir(&repo_path_str);
-    filesystem::create_dir(&repo_path_str);
+    // The dir might not exist which would throw an error. But here we don't care about that
```
denisidoro: Do you suggest another approach? I agree this is very error-prone

src/handler.rs
```diff
@@ -2,21 +2,40 @@ use crate::flows;
...
         None => flows::core::main(Variant::Core, config, true),
+
         Some(c) => match c {
             Preview { line } => flows::preview::main(&line[..]),
-            Query { query } => flows::query::main(query.clone(), config),
-            Best { query, args } => flows::best::main(query.clone(), args.to_vec(), config),
-            Search { query } => flows::search::main(query.clone(), config),
+
+            Query { query } => {
+                let error_string = format!("Failed to filter cheatsheets for {}", &query);
```
denisidoro: Wouldn't `with_context` prevent us from generating this str 100% of the times?

src/parser.rs
```diff
@@ -3,38 +3,80 @@ use crate::filesystem;
...
+                    "--preview-window" => opts.preview_window = Some(value.to_string()),
+                    "--header" => opts.header = Some(value.to_string()),
+                    "--overrides" => opts.overrides = Some(value.to_string()),
+                    _ => (),
+                }
+                Ok(())
+            } else if let [flag] = flag_and_value {
+                Err(anyhow!("No value provided for the flag `{}`", flag))
+            } else {
+                unreachable!() // Chunking by 2 allows only for tuples of 1 or 2 items...
```
denisidoro: Why not use an else in the line 72 instead?

src/structures/error/file_issue.rs
```diff
@@ -0,0 +1,23 @@
+use std::fmt::Debug;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+#[error(
+    "\rHey listen! Navi encountered a problem.
```
denisidoro: 😂

src/flows/repo.rs
Kibouo: Rust requires you to use the return value of `remove_dir`. But as stated in the comment above it, we don't care about the potential `Err` value. And the `Ok` type is `()` anyway.

src/parser.rs
Kibouo: The `chunks(2)` call on :46 returns slices (`&[..]`) of the collection. The problem is that, despite telling `chunks` that the slices should be `2` long, it doesn't know this at compile time. So Rust wants us to handle every possible case of slice length exhaustively.

Concrete this means that Rust wants the following let binding to be exhaustive.
```rust
} else {
    let [flag] = flag_and_value;
    Err(anyhow!("No value provided for the flag `{}`", flag))
}
```
image

You could do the following but obviously it wouldn't be as helpful an error:
```rust
} else {
    Err(anyhow!("No value provided for a flag"))
}
```

src/flows/repo.rs
Kibouo: Checking whether the dir exists before trying to remove it is a bit more robust, but not worth it imho. The check + remove is not atomic so `remove_dir` would still be able to return the same errors. Just less frequently.

src/handler.rs
Kibouo: `query` is part of `config`. But `config` is given away to `flows::query::main` on :17, which makes it impossible to use `query` after that. So we need to make a copy of `query` before that happens. 

`let query_clone = query.clone()` and then passing that to `with_context` would be a possibility. Not sure how smart the compiler is. I'll change it to be sure :thinking:

src/filesystem.rs
```diff
@@ -1,90 +1,122 @@
...
-    P: AsRef<Path>,
+    P: AsRef<Path> + Display,
 {
+    let error_string = format!("Failed to read lines from `{}`", filename);
     let file = File::open(filename)?;
-    Ok(io::BufReader::new(file).lines())
+    io::BufReader::new(file)
+        .lines()
+        .map(|line| line.map_err(Error::from))
+        .collect::<Result<Vec<String>, Error>>()
```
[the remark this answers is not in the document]
Kibouo: [In the `Err` case it short circuits.](https://doc.rust-lang.org/src/core/result.rs.html#1438) But best case it indeed allocates unnecessarily. Fixed!

PR #313, opened 2020-03-24
Fix path error on initial run

### LINKED ISSUE #312: Fix error reading paths in initial run
Capture d’écran 2020-03-24 à 11 36 33

When run for the first time, some paths don't exist.

That should be OK because we'll create them in a moment. But the way the code is at the moment, errors will short-circuit the flow.

denisidoro:
Fixes #312 

This isn't idiomatic but it'll do for the time being


denisidoro:
@Kibouo

Issue #227, opened 2020-03-12
Integration with Skim


ngirard:
[Skim](https://github.com/lotabout/skim) is a nice Rust equivalent for Fzf.

How about integrating Navi with Skim ? In could consist in either
- make it possible to call the `skim` binary as an alternative for `fzf` ;
- use Skim as a library.

denisidoro:
Duplicate #115

denisidoro:
In the past I tried to support skim but I gave up because it would require too many changes.

With the current implementation, this is the only file one needs to change: https://github.com/denisidoro/navi/blob/master/src/fzf.rs

denisidoro:
Related: https://github.com/lotabout/skim#use-as-a-library

PR #318, opened 2020-03-27
Add support for skim binary instead of fzf binary

shenek:
* added a compile time feature `use_skim` which causes that `sk` binary is used instead of `fzf`.
* refactored some code to be more `fzf/skim` neutral.

related issue #227

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,10 +14,6 @@ license = "Apache-2.0"
 [badges]
 travis-ci = { repository = "denisidoro/navi", branch = "master" }
 
-[features]
-default = []
-use_skim = []
-
 [dependencies]
 regex = "1.3.4"
 structopt = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,10 @@ license = "Apache-2.0"
 [badges]
 travis-ci = { repository = "denisidoro/navi", branch = "master" }
 
+[features]
+default = []
+use_skim = []
+
 [dependencies]
 regex = "1.3.4"
 structopt = "0.3"
--- a/src/structures/option.rs
+++ b/src/structures/option.rs
@@ -1,4 +1,14 @@
+use crate::finder::FinderChoice;
+use anyhow::Error;
 use structopt::{clap::AppSettings, StructOpt};
+
+fn parse_finder(src: &str) -> Result<FinderChoice, Error> {
+    match src {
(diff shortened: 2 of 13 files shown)


src/structures/option.rs
```diff
@@ -37,13 +37,13 @@ pub struct Config {
     #[structopt(long)]
     pub no_preview: bool,
 
-    /// FZF overrides for cheat selection
-    #[structopt(long, env = "NAVI_FZF_OVERRIDES")]
-    pub fzf_overrides: Option<String>,
+    /// finder overrides for cheat selection
+    #[structopt(long, env = "NAVI_FINDER_OVERRIDES")]
```
denisidoro: This is a breaking change :(

I think we should keep the old name even if that doesn't make much sense. At least for now

src/structures/option.rs
```diff
@@ -37,13 +37,13 @@ pub struct Config {
...
+    pub finder_overrides: Option<String>,
 
-    /// FZF overrides for variable selection
-    #[structopt(long, env = "NAVI_FZF_OVERRIDES_VAR")]
-    pub fzf_overrides_var: Option<String>,
+    /// finder overrides for variable selection
+    #[structopt(long, env = "NAVI_FINDER_OVERRIDES_VAR")]
```
denisidoro: Same as above

Cargo.toml
```diff
@@ -14,6 +14,10 @@ license = "Apache-2.0"
 [badges]
 travis-ci = { repository = "denisidoro/navi", branch = "master" }
 
+[features]
+default = []
+use_skim = []
```
denisidoro: Where is the skim dependency?

src/structures/option.rs
shenek: Ack. Reverted.

src/structures/option.rs
shenek: dtto

Cargo.toml
shenek: It is not dependent on skim during compilation. It just requires the binary. Works in the same way as fzf.

denisidoro:
Actually, since it uses the skim binary (instead of skim library) I see no reason to use compile-time checks for this.

I think we could have an option `--finder` (which fallbacks to `$NAVI_FINDER` which fallbacks to `fzf`)

I'll merge this on Monday, ok?

shenek:
I've done the changes required to pass `--finder` option. So feel free to check it and merge it.

PR #330, opened 2020-04-07
use zle redisplay

kaizhang91:
The original behavior:
1. The whole line in terminal will be killed
2. The new command will be printed one character by one character

The new behavior:
1. The selected command will be printed on terminal directly.

The code is based on <https://github.com/junegunn/fzf/wiki/examples#locate>.

### DIFF
--- a/shell/navi.plugin.zsh
+++ b/shell/navi.plugin.zsh
@@ -1,10 +1,11 @@
 #!/usr/bin/env zsh
 
 _call_navi() {
-   local -r buff="$BUFFER"
-   local -r r="$(printf "$(navi --print </dev/tty)")"
-   zle kill-whole-line
-   zle -U "${buff}${r}"
+  local selected
+  if selected="$(printf "$(navi --print </dev/tty)")"; then
+    LBUFFER="$selected"
+  fi
+  zle redisplay
 }
 
 zle -N _call_navi


shell/navi.plugin.zsh
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env zsh
 
 _call_navi() {
-   local -r buff="$BUFFER"
-   local -r r="$(printf "$(navi --print </dev/tty)")"
-   zle kill-whole-line
-   zle -U "${buff}${r}"
+  local selected
+  if selected=$(printf "$(navi --print </dev/tty)"); then
```
denisidoro, line 5: shouldn't we wrap `$(...)` with double quotes in order to include all words?

shell/navi.plugin.zsh
```diff
@@ -1,10 +1,11 @@
...
+    LBUFFER=$selected
```
denisidoro, line 6: Ditto

shell/navi.plugin.zsh
kaizhang91: You're right, and it's fixed.

shell/navi.plugin.zsh
kaizhang91: fixed

PR #336, opened 2020-04-11
Shell out to git clone

### LINKED ISSUE #305: Shell out to download repos
`navi repo add` uses [git2](https://docs.rs/git2/0.13.0/git2/) to clone repos. Some issues have arisen because of this. 

I plan to change this to curl + tar equivalents. I'll probably shell out because I don't want to add curl + tar + gzip dependencies + keep the OpenSSL one. 

### Pros
- cloning will be faster because it will be possible to download the master zip or tar.gz file directly
- smaller binary size
- it won't rely on local git configs

### Cons 
- if the machine running navi doesn't have canonical implementations of curl or tar, things will fail
- cloning private repos that require ssh won't work

I still haven't made my mind yet so if anyone has strong opinions about it, please shout out.

denisidoro:
Fixes #305

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,14 +67,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cc"
-version = "1.0.50"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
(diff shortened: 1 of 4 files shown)


ryuta69:
Oh I totally forgot to look up.
Thank you very much for fix.

PR #374, opened 2020-08-13
similar command line filter as for zsh

wardwouts:


denisidoro:
This won't work with bash < 4 (OSX's out-of-the-box one)

We'll need to have some fallback

wardwouts:
I did not realize that the bash version in OSX is that old. Bash 4 dates back to 2009.

PR #398, opened 2020-09-09
Revamp preview window for argument selection

### LINKED ISSUE #320: Show snippet in variable prompt
### Feature Request

The cheat information should/could stay on screen when choice is selected.

### Rationale

Example: I want to echo a variable, so I find the cheat:

image

I select the cheat. My screen shows this:

image

I get distracted by a colleague then look back, and I see the above. The screen has very little information about what is going on. I don't know what I am about to execute.

### Resolution

Can we (have an option to) make sure the cheat info is available on screen after selection?

Thank you for producing this tool, it looks great. I may even finally learn Rust for it.

denisidoro:
Fixes #320

Please let me know what you think

ezgif-4-d3992e2ee506


hlascelles:
That looks really good. Would love to give it live try. Will try a local build.

PR #436, opened 2020-11-24
Trigger edit on the current cheat file

### LINKED ISSUE #357: Triggering edit on the current cheat file.
Hi!

Is it possible to implement a way to trigger own editor on the current or even just a certain navi cheat file based on a keybinding?

Now feels quite cumbersome that it needs to separately navigate to those files before editing.

b3s1r:
The currently selected item can now be opened with ctrl-o in the standard editor defined by VISUAL or EDITOR.
fixes #357

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,6 +148,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "edit"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "323032447eba6f5aca88b46d6e7815151c16c53e4128569420c09d7840db3bfc"
+dependencies = [
+ "tempfile",
+ "which",
+]
+
+[[package]]
 name = "getrandom"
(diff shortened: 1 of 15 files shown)


denisidoro:
Cool! Thanks!

But I was wondering if we could have a cheaper solution: instead of appending the filepath to each line, we could append an index only and, when hitting Ctrl-O, we could match that with an array/map to get the filepath back. 

Please let me know if you intend to work in such an implementation. If not, I can merge it as-is and work on it later.

b3s1r:
Yeah, good idea!
I'll have a think about it and try to implement in the next few days.

denisidoro:
Thanks!

However, this way the `files` vec will contain a lot of repeated filepaths, right? We could reuse indices.

I'll merge it as-is and try to change this bit tomorrow.

denisidoro:
By the way, it won't work as expected if the snippet has a variable. I'll fix it as well.

Issue #449, opened 2020-12-29
Support for multiline formatted preview of code snippet


m42martin:
**navi --version 2.9.0
WSL2 on Windows 10 2004 19041.685
Ubuntu 20.04
Windows Terminal Preview**

I'd like be able to preview multiline snippets so that they appear as multiline, formatted, and possibly with syntax coloring related to the language.

I haven't found a way to display the code snippet as a multiline block in the preview.  If this is already possible, I would appreciate suggestions or an example illustrating how this should be configured.

![image](https://user-images.githubusercontent.com/16886205/103291982-e0d91580-49ba-11eb-8d9a-dd88fb9bb1d7.png)

Here's the content of the .cheat file that is selected in the image above:
```
% throw, expression, syntax, c#, cs, .net

# Throw expression within property

string Name \
{ \
	get name; \
	set => name = value \
	  ?? throw new ArgumentNullException(nameof(name)); \
}
```
navi arguments
```
navi --print \
 --finder 'skim' \
  --fzf-overrides '--nth 1,2,3 --no-exact --case smart --multi --height 15 --layout=reverse-l
ist --margin=2,4 --inline-info --ansi --preview-window=right:42%:wrap --with-nth 2,3'
```

I tried numerous ways of calling `skim` but none have resulted in a clean and formatted preview.

I thought it might be possible using [bat](https://github.com/sharkdp/bat) or [prettybat](https://github.com/eth-p/bat-extras/blob/master/doc/prettybat.md) with skim's `--preview` option.

e.g., `--preview bat {3}` 

![image](https://user-images.githubusercontent.com/16886205/103293254-7ecddf80-49bd-11eb-8031-857e0870a599.png)

I've seen `: command not found` in the preview output from many variations that I've tried.

When wrapping the preview command in quotes, or providing inline arguments to a command, other issues arise

`--preview "prettybat {3}"` or `--preview "prettybat --language cs {3}"`
```
/bin/bash: -c: line 0: unexpected EOF while looking for matching `"'                                               /bin/bash: -c: line 1: syntax error: unexpected end of file
```

While this works by piping the cheat file into awk, replacing `\` with `\n`
```
awk '{ gsub(/\\/, "\n", $0); print }'  file.cheat
```
Doing the same within the fzf-overrides results in the 'unexpected EOL' error

 `--preview awk "{ gsub(/\\/, "\n", {3}); print $0; }"'` 

[...]

PR #450, opened 2020-12-29
Proof of concept for code snippets

denisidoro:
Screen Shot 2020-12-29 at 17 55 06

### DIFF
--- a/src/cmds/core.rs
+++ b/src/cmds/core.rs
@@ -271,7 +271,7 @@ pub fn main(config: Config) -> Result<(), Error> {
     env::set_var(env_vars::PREVIEW_COMMENT, &comment);
 
     let interpolated_snippet = display::with_new_lines(
-        replace_variables_from_snippet(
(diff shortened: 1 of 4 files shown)


enisozgen:
Hello Dennis ,

Do you think to implement this to navi?

I also read the https://github.com/denisidoro/navi/issues/449.  

In my opinion, this feature request is a bit overkill(syntax highlight for language etc), it will change the concept so much.

Regards,
Enis

denisidoro:
I don't think I will. I honestly believe there are better tools out there for this purpose.

ngirard:
@denisidoro, @engrravijain, which tools would you folks recommend for this ? Are you using one ? If so, how does it benefit you ?

Thanks in advance !

ngirard:
> which tools would you folks recommend for this ?

Replying to myself: this is [the way](https://github.com/out-of-cheese-error/the-way).

aemonge:
Still off road map?

PR #490, opened 2021-04-10
Publish binaries for more platforms, including Windows

denisidoro:


ngirard:
[Watchexec](https://github.com/watchexec/watchexec) has a nice [release.yml](https://github.com/watchexec/watchexec/blob/main/.github/workflows/release.yml) that deals with builds on Windows -- and incidentally producing `.deb` packages, which is appreciable.

Hope that helps

PR #560, opened 2021-06-01
Rephrase etymology to active voice

pbsds:
Also adds the "navigator" origin of Navi


denisidoro:
I'm OK with the active voice but originally navi isn't named after `Navigator`: it's named after `Navi` (the character) which in turn is probably named after `Navigator`.

Could you remove this part, please?
```
short for "Navigator"
```

pbsds:
Just squash the two commits ;)

PR #670, opened 2022-01-13
Update legacy syntax from Elvish widget

cherryblossom000:
This PR updates the following legacy syntax from the Elvish widget so it works on v0.17 without deprecation errors:

- assignment: `foo = bar` → `var foo = bar` or `set foo = bar`
- lambdas: `[arg1 arg2]{ body }` → `{|arg1 arg2| body }`; all the lambdas in the widget have no arguments so they can simply be written as `{ body }` instead of `{|| body }` which coincidentally keeps the widget compatible with pre-0.17 versions

This PR also refactors some of the code so it's cleaner.

---

As the `var` and `set` special commands were introduced in Elvish v0.15 (January 2021), this updated Elvish widget will only work in versions 0.15 and up, so this PR is technically a breaking change.

### DIFF
--- a/shell/navi.plugin.elv
+++ b/shell/navi.plugin.elv
@@ -1,31 +1,29 @@
 use str
 
-fn call-navi []{
+fn call-navi {
   if (eq $edit:current-command '') {
-    answer = (navi --print)
+    var answer = (navi --print)
     edit:replace-input $answer
   } elif (not (str:contains-any $edit:current-command '|')) {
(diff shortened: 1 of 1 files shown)


denisidoro:
Even though this is a breaking change, the minimum supported version is 1+ year old.

I'll assume that if a given user hasn't updated elvish in a while, they won't update navi either, so it's OK.

PR #678, opened 2022-01-27
Add build feature to disable command execution

ukautz:
This CL adds the build feature `disable-command-execution` that enforces the `--print` flag and thereby disables execution of commands. Basically `--print` is enforced and cannot be disabled.

I think this proposed change will make `navi` usable in a wider range of situations, especially when a single execution can have far reaching impacts. Think: operations.

**Rationale:**
`navi` is great, because it allows users, who are not deeply familiar with a tooling, to still use it at an expert level: The underlying cheatsheets are persisted, combined knowledge; the variables add environment state. Both together make it a powerful idea.
Going back to above scenarios, in which a single keystroke can have great consequences: Extra care of executing the generated commands should be taken. This extra care is far more likely to take place when a manual component is involved, i.e: if you have to copy & paste the command and then execute it yourself, you are more likely to take responsibility and therefore care.

I understand, that `navi` also makes great sense in other scenarios where this additional care is not needed / wanted, hence the proposal of a build flag, to make this tool available to either one.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,9 @@ keywords = ["cheatsheets", "terminal", "cli", "tui", "shell"]
 categories = ["command-line-utilities"]
 license = "Apache-2.0"
 
+[features]
+disable-command-execution = []
+
 [badges]
 travis-ci = { repository = "denisidoro/navi", branch = "master" }
 
--- a/src/config/cli.rs
+++ b/src/config/cli.rs
@@ -97,6 +97,7 @@ pub(super) struct ClapConfig {
 
     /// Instead of executing a snippet, prints it to stdout
     #[clap(long)]
+    #[cfg(not(feature = "disable-command-execution"))]
     pub print: bool,
 
     /// Returns the best match
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -136,8 +136,18 @@ impl Config {
         self.yaml.style.comment.min_width
     }
 
+    #[cfg(feature = "disable-command-execution")]
+    fn print(&self) -> bool {
+        true
+    }
+
+    #[cfg(not(feature = "disable-command-execution"))]
+    fn print(&self) -> bool {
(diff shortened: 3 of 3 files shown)


denisidoro:
Nice. 

But there are a lot of other shelling out spread across the code base which isn't covered by this change. Were you aware of it?

In fact, there's some shelling out that is critical for navi to work, such as invoking fzf. So I think there will always be some way to circumvent navi and execute arbitrary code

denisidoro:
Closed by mistake

denisidoro:
I'm OK with merging it as-is. I just want to confirm if your expectations are aligned

ukautz:
> In fact, there's some shelling out that is critical for navi to work, such as invoking fzf. So I think there will always be some clever way to circumvent navi and execute arbitrary code

Thanks for mentioning it. That's perfectly fine with me :) My concern, this PR addresses, is not about calling external tooling per-se. It's really just about not defaulting to call the command `navi` found, without having the user review it.

I really love how `navi` makes complicated tooling accessible for users, who are not extremely familiar with it. And in those cases, especially if the tooling can have far reaching consequences, I just think that extra layer of care (having to copy & paste & enter) should be the default. In all other scenarios execution of the found command is likely the better choice. With the build flag we can have both :)

aemonge:
Can this be set at settings? I want to default `--print` and only execute if `navi -x`

denisidoro:
You can write a wrapper in your shell for this.

```sh
mynavi() {
  if...  then
   navi --print... 
  else
   navi... 
   fi
}
```

I'm on my phone right now but hopefully you get the idea

aemonge:
I get the idea, through aliasing, but I that feels temporary thought. 

I mean, for my sake, I do alias it as suggested with simpler `alias navi='navi --print'` , but would be nice to read all parameter through config, and overwritten them if present as parameter; as common practice :)

PR #686, opened 2022-02-11
Add shared paths for cheat directory and config file

ukautz:
This CL makes it easier to install navi on multi-tenant Linux systems, by allowing to provide pre-install cheatsheets in shared folders (and potentially pre-installed config in shared folders).

This will align navi with "other standard tooling", that allow you to set a path in the user folder, but use a shared (system) folder if that is not specified.

Note: This does not break existing setups. The shared cheats directory / config file is only used if:
- There is not already a directory / config file in the user XDG location
- Navi is compiled with `NAVI_CONFIG_FILE` and/or `NAVI_CHEATS_DIRECTORY` env var set
- The paths defined in above env vars exist

### DIFF
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -9,7 +9,7 @@ use anyhow::Result;
 use directories_next::BaseDirs;
 use regex::Regex;
 use std::collections::HashSet;
-use std::path::{Path, PathBuf};
+use std::path::{Path, PathBuf, MAIN_SEPARATOR};
 use walkdir::WalkDir;
 
 pub fn all_cheat_files(path: &Path) -> Vec<String> {
@@ -29,6 +29,11 @@ fn paths_from_path_param(env_var: &str) -> impl Iterator<Item = &str> {
 fn compiled_default_path(path: Option<&str>) -> Option<PathBuf> {
     match path {
         Some(path) => {
+            let path = if path.contains(MAIN_SEPARATOR) {
+                path.split(MAIN_SEPARATOR).next().unwrap()
+            } else {
+                path
+            };
             let path = Path::new(path);
             if path.exists() {
                 Some(path.to_path_buf())
@@ -46,7 +51,7 @@ pub fn default_cheat_pathbuf() -> Result<PathBuf> {
     pathbuf.push("navi");
     pathbuf.push("cheats");
     if !pathbuf.exists() {
-        if let Some(path) = compiled_default_path(option_env!("NAVI_CHEATS_DIRECTORY")) {
+        if let Some(path) = compiled_default_path(option_env!("NAVI_PATH")) {
             pathbuf = path;
         }
     }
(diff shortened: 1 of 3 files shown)


src/filesystem.rs
```diff
@@ -41,6 +59,11 @@ pub fn default_config_pathbuf() -> Result<PathBuf> {
     let mut pathbuf = PathBuf::from(base_dirs.config_dir());
     pathbuf.push("navi");
     pathbuf.push("config.yaml");
+    if !pathbuf.exists() {
+        if let Some(path) = compiled_default_path(option_env!("NAVI_CONFIG_FILE")) {
```
denisidoro, line 63: Instead of creating a new env var, could we please reuse https://github.com/denisidoro/navi/blob/8725654a746fa8c999fbd5437559fdfcf1f7fb53/src/env_var.rs#L19?

src/filesystem.rs
```diff
@@ -26,12 +26,30 @@ fn paths_from_path_param(env_var: &str) -> impl Iterator<Item = &str> {
...
+}
+
 pub fn default_cheat_pathbuf() -> Result<PathBuf> {
     let base_dirs = BaseDirs::new().ok_or_else(|| anyhow!("Unable to get base dirs"))?;
-
     let mut pathbuf = PathBuf::from(base_dirs.data_dir());
     pathbuf.push("navi");
     pathbuf.push("cheats");
+    if !pathbuf.exists() {
+        if let Some(path) = compiled_default_path(option_env!("NAVI_CHEATS_DIRECTORY")) {
```
denisidoro, line 49: Instead of creating a new env var, could we please reuse https://github.com/denisidoro/navi/blob/8725654a746fa8c999fbd5437559fdfcf1f7fb53/src/env_var.rs#L14?

We would need to split it by `:` and get the first element.

denisidoro (changes requested):
Could you also please add some brief documentation about it in `/docs`? Thanks!

src/filesystem.rs
ukautz: Hi @denisidoro. Certainly, happy to replace both env var names with the existing ones.

Just to be sure: both existing env vars are currently used at *runtime* to override / set the respective paths for navi to look into. This CL would use them at  *compile time* for the slightly different purpose of setting a fallback for the respective paths, in case they do not exist (i.e. if there is a system wide installation then navi will use that - if the users has not decided to set up a custom one in their home directory). So the env vars would be a bit "dual purpose".

src/filesystem.rs
```diff
@@ -26,12 +26,35 @@ fn paths_from_path_param(env_var: &str) -> impl Iterator<Item = &str> {
     env_var.split(':').filter(|folder| folder != &"")
 }
 
+fn compiled_default_path(path: Option<&str>) -> Option<PathBuf> {
+    match path {
+        Some(path) => {
+            let path = if path.contains(MAIN_SEPARATOR) {
+                path.split(MAIN_SEPARATOR).next().unwrap()
```
jirutka, line 33: What is this supposed to do? It doesn’t work correctly, `"/etc/xdg/navi/config.yaml"` is transformed to `""`.

PR #776, opened 2022-08-11
Fix preview map "leaking" with multiple variables

trantor:
With multiple variables in the preview, each with
different «--map» options, there is an application
of the map code for the current variable to the
preview of the other variables as well.
This code fixes that.

The problem is limited only to the preview.

NOTE: @denisidoro this is my first time handling rust code, so any correction is welcome.


trantor:
@denisidoro what do you think?

PR #833, opened 2023-04-22
Use XDG conventions on macOS too

utkarshgupta137:
`~/Library/Application Support` is an utterly inconvenient location for macOS users. Most CLI tools follow the `XDG conventions` on macOS as well. Replace the `directories-next` crate with the `etcetera` crate to use it on macOS. Config locations for other platforms are unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -282,6 +282,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "etcetera"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "51822eedc6129d8c4d96cec86d56b785e983f943c9ce9fb892e0c2a99a7f47a0"
+dependencies = [
+ "cfg-if",
+ "home",
+]
+
+[[package]]
 name = "fs_at"
 version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


denisidoro:
Isn't this a breaking change? What will happen to existing OSX users?

utkarshgupta137:
> Isn't this a breaking change? What will happen to existing OSX users?

Right now this PR makes no effort to maintain backward compatibility on macOS. IMO, no macOS users expect the config file to be located in `~/Library/Application Support`. Almost every CLI tool which uses the `dirs`/`dirs-next` crates & other clones has open or closed issues of users wanting to use the XDG conventions on macOS. But I've yet to come across a case where the inverse is true. I've written more about this [here](https://github.com/dirs-dev/directories-rs/issues/47#issuecomment-1464129466).
However, since you probably want to maintain backward compatibility, I can add migration code to move the old data to the new location, similar to this: https://github.com/cantino/mcfly/pull/349/files#diff-4f91d834bba4a3a432715f3ccf4ebc7c10d7ecfb974f1e78f56108dd80581a53R392-R403

denisidoro:
> However, since you probably want to maintain backward compatibility, I can add migration code to move the old data to the new location

That would be awesome!

Perhaps a simpler approach could be the following?
1. if on osx, try to use ~/Library/.../navi/... if it exists
2. otherwise, use whatever etcetera determines

utkarshgupta137:
> > However, since you probably want to maintain backward compatibility, I can add migration code to move the old data to the new location
> 
> That would be awesome!
> 
> Perhaps a simpler approach could be the following?
> 
>     1. if on osx, try to use ~/Library/.../navi/... if it exists
> 
>     2. otherwise, use whatever etcetera determines

Done.

utkarshgupta137:
@denisidoro Waiting for a release :)

Issue #576, opened 2021-07-06
Nothing of cheats available on Windows


acodecow:
**Describe the bug**
navi doesn't not recognize any cheat sheets.

**To Reproduce**
run navi, if *current working directory* is not at `C:/` driver,

**Expected behavior**
It should **NOT REWRITE** disk driver character when query profile data

**Screenshots**

![image](https://user-images.githubusercontent.com/13063376/124584523-edb55500-de86-11eb-987d-e17c2bda7f81.png)

**Versions:**
 - OS: Windows
 - using any windows shell: navi 2.16.0

**Additional context**
sceenshot from `procmon` capture.

denisidoro:
Thanks for troubleshooting, but I'm afraid I didn't get it.

Do you mean that `navi.exe` should `CreateFile` in `C:\Users\...\cheats` but instead it's trying to create it in `D:\Users\...\cheats` which is invalid?

LittleboyHarry:
![image](https://user-images.githubusercontent.com/13063376/124614218-9aeb9580-dea6-11eb-8b3d-b295d1cc7578.png)

denisidoro:
I believe this is a bug in https://docs.rs/directories-next/2.0.0/directories_next/struct.ProjectDirs.html#method.config_dir

I'll need to investigate this upstream

scillidan:
I also had the similar problem.
I run it in `Cmder`.
![image](https://user-images.githubusercontent.com/19889610/185138892-5efaf3d4-1fd2-4d7c-8424-9048267b038a.png)

phanirithvij:
@scillidan I think you need to set cmd.exe as the shell in navi config. ref: https://github.com/denisidoro/navi/issues/40#issuecomment-845495970

zjp-CN:
This is caused by using `:` in path join https://github.com/denisidoro/navi/blob/6f316ec7eb2f1f56a7609048e67860eb84c1aacf/src/config/mod.rs#L72

https://github.com/denisidoro/navi/blob/6f316ec7eb2f1f56a7609048e67860eb84c1aacf/src/filesystem.rs#L155

https://github.com/denisidoro/navi/blob/6f316ec7eb2f1f56a7609048e67860eb84c1aacf/src/filesystem.rs#L25-L27

It's incorrect since `:` is a valid component on Windows. It'd be better to use PathBuf instead of String for path and carry the vec of path along.

[...]

PR #840, opened 2023-05-13
Fix multiple paths: define the platform-specific join separator

zjp-CN:
fix https://github.com/denisidoro/navi/issues/576

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 **/*.rs.bk
+navi.log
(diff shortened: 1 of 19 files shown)


zjp-CN:
But the preview is broken too in debug mode. 




The path is not right.


I think it's related to https://github.com/rust-lang/rust/issues/99931 and https://github.com/rust-lang/rust/issues/59117

denisidoro:
@zjp-CN sorry for taking so long to review this. I'm not able to test this on a Windows machine. Is this a change you're comfortable with landing?

jackandroselv:
您好！您的邮件我已收到，我会尽快的查阅邮件。谢谢您的来信！

zjp-CN:
> Is this a change you're comfortable with landing?

Yes, I use the compiled navi based on this PR on Windows all the time, and it works fine.

This PR mainly fixes the path parsing on Windows (so nothing changes on Linux), and introduces tracing for logging.

PR #853, opened 2023-08-21
Update badge link for CI workflow

kachick:
Looks like the workflow has been renamed since 2ca1d8fc85333c1a8414a71e288073c264d74b1f

## Before

before

## After

after


kachick:
Thank you!

Issue #777, opened 2022-08-12
Installing navi correctly on MacOS for .zsh


AnvarAtayev:
I am just wondering if I have installed navi correctly.

I am using macOS Monterey 12.5 and installed navi using homebrew via `brew install navi`.
To install the shell widget, I pasted the following into the end of my `.zshrc` file:
```
# zsh
eval "$(navi widget zsh)"
```
(I do not have a `.bshrc` file, and thought this was the alternative).

Having done the above step, I am now able to use navi by calling Ctrl + G as expected.

Entering `navi fn welcome` and entering the `config` command, I get a permission denied error.
<img width="878" alt="Screenshot 2022-08-12 at 11 30 52" src="https://user-images.githubusercontent.com/70859208/184336904-a0b1f0ce-c507-49ff-8385-f9e7cd915ec1.png">
<img width="878" alt="Screenshot 2022-08-12 at 11 32 23" src="https://user-images.githubusercontent.com/70859208/184337111-53ed8950-5110-405f-adc6-df8255660c87.png">

I'm not sure how to overcome this. Also, I am noticing that it is trying to use `bash` instead of `zsh`. Not sure how to change this also.

Any help would be appreciated.

wiegell:
same on armbian

jjournet:
I have the same issue on MacOS Monterey (M1), and I think the problem is related to the hardcoded path of the config file containing a space, and not handled correctly by navi
2 things:

- it would be better for navi to rely on a default config file system-wide when no config at user level (so that we can change the location of the config file), for instance in /etc/navi
- the default config file should be in ~/.config/navi , not in Library (IMHO)

jjournet:
I would say issue was introduced in change #518 
I don't code in Rust, so I can't propose a fix

jjournet:
Actually, in the current version, it's not working on Mac, because you can't have a config file (Navi doesn't find it because of the space), the checkout of new libraries works (I was able to get the defaults cheats), but navi doesn't see them

dmartzol:
> the default config file should be in `~/.config/navi`, not in Library (IMHO)

+1

Most CLI apps have the config in `~/.config`. Having the config file in a different location just makes dotfiles management more difficult, IMHO

denisidoro:
The config dir is defined by https://crates.io/crates/directories-next

Changing it now would be a breaking change :/

denisidoro:
I'm trying to understand if the problem is really with the space. Could you provide more logs / evidence?

PR #859, opened 2023-11-04
docs: clarify directory defaults and environment variables

tjex:
There are a few questions (#777, #644 , and no doubt a few others) that stem from the confusion of default directories (particularly on mac) and how to set custom destinations. 

This documentation addition hopefully clears that up sufficiently.


tjex:
@denisidoro ?

This pr could well mitigate a lot of recurrent user issues.

PR #898, opened 2024-05-29
Update libc & linux-raw-sys to support loongarch64

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.
Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html
Due to the lower version of libc & linux-raw-sys, an error occurs when compiling on the loongarch64 architecture.
For example:
```
error[E0432]: unresolved import `crate::general`
  --> /home/alpine/.cargo/registry/src/index.crates.io-6f17d22bba15001f/linux-raw-sys-0.3.1/src/lib.rs:84:16
   |
84 |     use crate::general::{cmsghdr, msghdr};
   |                ^^^^^^^ could not find `general` in the crate root
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -397,15 +397,15 @@ checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
 
 [[package]]
 name = "libc"
-version = "0.2.140"
+version = "0.2.155"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


jackandroselv:
您好！您的邮件我已收到，我会尽快的查阅邮件。谢谢您的来信！

PR #915, opened 2024-09-16
Change default delimiter

kit494way:
Currently, the default delimiter is **two** or more white space characters.
The default delimiter should be **one** or more white space characters.

### DIFF
--- a/src/finder/post.rs
+++ b/src/finder/post.rs
@@ -43,7 +43,7 @@ _navi_input | _navi_map_fn | _navi_nonewline"#
 fn get_column(text: String, column: Option<u8>, delimiter: Option<&str>) -> String {
     if let Some(c) = column {
         let mut result = String::from("");
-        let re = regex::Regex::new(delimiter.unwrap_or(r"\s\s+")).expect("Invalid regex");
+        let re = regex::Regex::new(delimiter.unwrap_or(r"\s+")).expect("Invalid regex");
         for line in text.split('\n') {
             if (line).is_empty() {
                 continue;


denisidoro:
Won't this break commands with a single space?

kit494way:
After re-reading the code, I realized that this change could affect more than just the intended part.
My intention is to omit redundant `--delimiter` options for variables.
My use cases are here.

```
% gh

# download release asset
gh release download <release_tag> -R <release_repo> -p <asset_pattern>

$ release_tag: gh release list -R <release_repo> --exclude-pre-releases --exclude-drafts --json tagName,publishedAt --jq '.[] | [.tagName, .publishedAt] | @tsv' --- --column 1 --delimiter '\t'
$ asset_pattern: gh release view -R <release_repo> --json assets --jq '.assets[].name'
```

```
% mise

# use local
mise use <tool>@<version>

$ tool: mise plugins ls -c --user -q
$ version: join -a 1 <(mise ls-remote <tool> | sort) <(mise ls -i <tool> | awk '{print $2 "\t*"}' | sort) | sort -k 1 -Vr --- --column 1 --delimiter "\s+"
```

The first time I tried to use the `--column` option without the `--delimiter`, I was confused.
This is because I expected the default delimiter to be something like `\s+`.
In my opinion, `\s+` is more common than `\s\s+` as the delimiter.
Currently, when I use the `--column` option, I always specify the `--delimiter`, which I find redundant.

After reconsideration, I think the following change would better fit the intent of changing the default delimiter for variables.

```diff
diff --git a/src/finder/structures.rs b/src/finder/structures.rs
index a055194..231ff06 100644
--- a/src/finder/structures.rs
+++ b/src/finder/structures.rs
@@ -66,6 +66,7 @@ impl Opts {
             overrides: CONFIG.fzf_overrides_var(),
             suggestion_type: SuggestionType::SingleRecommendation,
             prevent_select1: false,
+            delimiter: Some(r"\s+".to_string()),
             ..Default::default()
         }
     }
```

Is this OK?

alexis-opolka:
@denisidoro @kit494way Is this PR still open to be merged? Or is there any change still pending?

denisidoro:
I'm afraid of the (potentially undesired) consequences of this change.

If there's evidence it's a safe change, I can approve it.

kit494way:
In the case of selecting a column that contains a single space without specifying a delimiter, this change will change the behavior.
I don't know if there is a specific case for that.

Instead of changing the default delimiter, can you approve a change that allows the default delimiter to be changed in a configuration file?
If so, I will close this pull request and create a new issue.

alexis-opolka:
@kit494way I think it is the best approach considering the possible consequences of changing the default behaviour of navi.
It seems denisidoro also agrees with creating a new configuration variable and make it a kind of opt-in behaviour for the users.

Let me know when your changes are to be merged into the main branch as I will take this opportunity to update the documentation on this new configuration variable in the same PR.

Good luck!

kit494way:
I opened a new issue #940 .

Issue #851, opened 2023-08-04
Please make navi info better


Apromixately:
**Is your feature request related to a problem? Please describe.**
`.oO( Where are my cheats? )`
`$ navi --help`
`.oO( no, that's not it... ) *scroll for a while*`
`.oO( ooooh, navi info, that was it )`
`$ navi info`
```
error: the following required arguments were not provided:
  <INFO>

Usage: navi info <INFO>

For more information, try '--help'.
```
(╯°□°）╯︵ ┻━┻

**Describe the solution you'd like**
Print the cheats-path and config-path when running navi info without arguments.

alexis-opolka:
I'm sorry for the long time you had to wait @Apromixately but while I'm making fixes for the `navi info` command with the PR #938, I'd like to know how you would like the command to behave.

I understood that you wanted it to display the `cheats-path` and the `config-path` when ran with no arguments, but I think it would be better to just display the help as the default behaviour.

An example of the output with the help included:

```bash
error: the following required arguments were not provided:
  <INFO>

Usage: navi.exe info <INFO>

Arguments:
  <INFO>  [possible values: cheats-example, cheats-path, config-path, config-example]

For more information, try 'navi info --help'.
```

It seems to me there is work to be done for users to know and understand where to get their cheats and config paths used by Navi.
I think the base `--help` command should have a bit more informations about what the info command does and maybe a better entry in navi's documentation.

Would that be good for you? Would you need anything else?

Issue #907, opened 2024-07-29
Base navi directory for cheats-path different from config-path


parkero:
**Describe the bug**
On macOS the base directory for cheats and config are different without `NAVI_*` environmental variables being present.

**To Reproduce**
```
% echo $(navi info config-path) 
/Users/parkero/.config/navi/config.yaml
```

```
% echo $(navi info cheats-path)`
/Users/parkero/Library/Application Support/navi/cheats
```

**Expected behavior**
Base navi path would be the same for both in the absence of user configuration

**Versions:**
- Navi: 2.23.0
- macOS:  13.6.6
- zsh 5.9 (x86_64-apple-darwin22.0)

**Additional context**
This came to my attention as I had a previous config.yaml changing my colors, but when I went to edit the config (through the default navi cheat "edit config file") it showed an empty file. 

Either location seems fine, but odd to have them spread throughout the file tree.

I can write up a pull request if I know what the desired behavior should be, or feel free to close if working as intended.

alexis-opolka:
So @denisidoro, what do we do regarding this issue?
Do we consider keeping the current paths? Or should we change the current behaviour in a later major release?
If we decide to change the behaviour, should we change it for all platforms in the same way or should we look into more platform/distribution -specific paths?

alexis-opolka:
@parkero I'll take a look at it and let you know in the following weeks, thanks.

PR #938, opened 2025-02-04
Add two distinctive info commands

alexis-opolka:
Modified info commands in a way where the value computed during compilation is prefixed with the `default` keyword and the command giving the current value doesn't have any prefix.

It should be able to remove any kind of confusion related to this command.
It can also help with debugging issues related to those variables.

I also updated the corresponding documentation.

Fixes: #851

---

Current tasks remaining before this draft can become a proper PR:

- [x] Handle Apple paths
- [x] Update current commands documentation


denisidoro:
Isn't this a breaking change? What is someone is relying on the old way and then updates navi? Will their setup continue working?

alexis-opolka:
Technically speaking, **it is a breaking change** since the behaviour of commands would change with the API.
In the case of users, it wouldn't change a lot I think since the old behaviour gave the default path and not the current path used by navi they would now either have the same path with both commands or finally have their scripts working with navi.

For mac users it would mean to move their cheatsheets to the same path but I think it is necessary to standardize the paths for all platforms and not have only one platform have two different paths for their configuration and their cheatsheets. So we should either have all the platforms have one path for their configuration and one path for their cheats or none at all.

There was also quite a lot of confusion about what was printed with `info config-path` and `info cheats-path` so I think this is for the better to finally have a `info default-*` command for the default values and a `info *` command for the values at execution.

kit494way:
There are other options to configure the path of cheat files.
- `--path` command line argument
- `cheats.paths` in config.yaml
- `cheats.path` in config.yaml, although this is deprecated

It is better to use `path()` method to get the cheat files path.
https://github.com/denisidoro/navi/blob/90a29ed3249073eb6495922999e942e9a22461a8/src/config/mod.rs#L63-L77

---

The configuration is also read from `NAVI_CONFIG_YAML` environment variable.
The method to get the configuration is defined here.
https://github.com/denisidoro/navi/blob/90a29ed3249073eb6495922999e942e9a22461a8/src/config/yaml.rs#L111-L126

But it seems there is no way to determine where the configuration was read from.
It might be better to add a field to `YamlConfig` that holds the source from which the configuration is read and display it.

alexis-opolka:
Oh, I didn't see that, thanks!
I'll try to make the necessary modifications in this week.

alexis-opolka:
While I'm working to take into account your changes @kit494way and I'm waiting an answer from #851, I'm moving this PR back as a draft, just to ensure that changes are not merged without all necessary modifications.

alexis-opolka:
@denisidoro @kit494way, I think I found something acceptable for the `config-path` and `cheats-path` commands.

- For the `cheats-path` command there is:

  ```rust
  pub fn current_cheat_pathbuf() -> Result<String> {
      let cheats = CONFIG.path();
  
      // We check if `Config.path()` returned a value
      if cheats.is_some(){
          // A value has been returned => We send it back
          Ok(cheats.unwrap().to_string())
      } else {
          // No value has been returned => no user-defined path has been found
          // We use the default idiomatic path for each supported platform.
          Ok(default_cheat_pathbuf()?.to_string())
      }
  }
  ```
  
  which should either return the currently used path/paths for the cheats or the default path defined at compile time.

  Outputs:
    
    - Cheats paths returned without any user-defined paths or input
    
      ```bash
      ~/navi> navi info cheats-path
      /home/test/.local/share/navi/cheats
      ```
    
    - Cheats paths defined in the default configuration file using `cheats.paths`
    
      ```bash
      ~/navi> navi info cheats-path
      ~/test/:~/another/dir/
      ```
    
    - Cheats paths defined in the default configuration file using the now deprecated property `cheats.path`
    
      ```bash
      ~/navi> navi info cheats-path
      my/path/now/deprecated/
      ```
    
    - Cheat paths defined using the `NAVI_PATH` environment variable
    
      ```bash
      ~/navi> NAVI_PATH="/my/path/with/env/" navi info cheats-path
      /my/path/with/env/
      ```
    
    - Cheat paths defined using the `--path` argument
    
      ```bash
      ~/navi> navi --path /my/path/with/arg/ info cheats-path
      /my/path/with/arg/
      ```

- For the `config-path` command, it's a bit more tricky but the base method is:

  ```rust
  pub fn current_config_pathbuf() -> Result<String> {
      Ok(CONFIG.get_source())
  }
  ```
  
  The `Config.get_source()` method serves only as a public way to access the `.source` attribute of the `YamlConfig` (as `Config.yaml` is private) which essentially is defined during the `YamlConfig.get()` method:
  
  ```rust
  pub fn get(env: &EnvConfig) -> Result<YamlConfig> {
      if let Some(yaml) = env.config_yaml.as_ref() {
          // We're getting the configuration from the environment variable `NAVI_CONFIG_YAML`
          let mut cfg = Self::from_str(yaml)?;
          cfg.source = format!("NAVI_CONFIG_YAML={}", yaml);
  
          return Ok(cfg);
      }
      if let Some(path_str) = env.config_path.as_ref() {
          // We're getting the configuration from a file given in the environment variable `NAVI_CONFIG`
  
          let p = PathBuf::from(path_str);
          let mut cfg = YamlConfig::from_path(&p)?;
          cfg.source = format!("NAVI_CONFIG={:?}", p).to_string();
  
          return Ok(cfg);
      }
      if let Ok(p) = default_config_pathbuf() {
          // We're getting the configuration from the default path
  
          if p.exists() {
              let mut cfg = YamlConfig::from_path(&p)?;
              cfg.source = p.to_string();
  
              return Ok(cfg);
          }
      }
  
      // As no configuration has been found, we set the YAML configuration
      // to be its default (built-in) value.
      Ok(YamlConfig::default())
  }
  ```
  
  This method has 4 possible outputs:
  
  - The configuration is defined with the `NAVI_CONFIG_YAML` environment variable => the `source` attribute equals to `NAVI_CONFIG_YAML=<the-yaml-config-string>`.
  
    > **Note**
    > Currently, the YAML string is printed as-is but it won't stay like this, I'm still thinking about how to print it in a useful way.
  
  - The configuration is defined from a file given with the `NAVI_CONFIG` environment variable => the `source` attribute equals to `NAVI_CONFIG=<path>`.
  - The configuration is defined from the default file present in the default path => the `source` attribute equals to `<path>`.
  - No configuration has been found and the built-in default configuration value is used => the `source` attribute equals to `built-in`.
  
  Outputs:

  ```bash
  ~/navi> navi info config-path
  built-in
  ```

  ```bash
  ~/navi> NAVI_CONFIG=/home/test/config.yaml navi info config-path
  NAVI_CONFIG="/home/test/config.yaml"
  ```

  ```bash
  ~/navi> NAVI_CONFIG_YAML="cheats:" navi info config-path
  NAVI_CONFIG_YAML=cheats:
  ```

  ```bash
  ~/navi> navi info config-path
  /home/test/.config/navi/config.yaml
  ```

What do you think about it? Did I miss something or is it good for me to make the finishing touches and the documentation entry?

alexis-opolka:
> When a config file does not exists at the default config path, it can be assumed that an empty file is read. In this case, it is better to display the default config path instead of `built-in` so that users can find out where to put a config file.

Is it really interesting though? Since `default-config-path` already points to the default path which makes it a reference for users to find where to put the config file.
I also think `config-path` can be used as a first way of debugging which configuration is in effect within navi's execution.

I think it would be weird to have it point to the default config path when it's not reading from it. How can we be sure then that the changes made inside the configuration file are read but not used if the only way to know if navi uses the file or not is `config-path` (for a user) apart from the not-so-easy-to-remember `RUST_BACKTRACE=rust` environment variable?

I already had to check which configuration was used while not editing the source code and I still haven't found the `navi.log` file in my filesystem when no configuration file was found and the default configuration (built-in) was used instead.

Honestly, if you have a better method for debugging navi as a user (which can't edit the source code) I would greatly appreciate it.

kit494way:
> I think it would be weird to have it point to the default config path when it's not reading from it. How can we be sure then that the changes made inside the configuration file are used by navi if the only way to know if navi uses the file or not is config-path (for a user) apart from the not-so-easy-to-remember RUST_BACKTRACE=rust environment variable?

Even if `config-path` outputs a file path, it does not mean that changes made in the file have been used.
When debugging configuration, we should check two points:
- whether the config file path is correct or not
- whether settings in the file is correct or not (e.g. Is there any typo?)

`config-path` is useful at the first point, but not at the second.

If there is a typo in the config file and navi uses default configuration, regardless of whether `config-path` shows config file path or “built-in”, user can not find out whether changes in the config file is used or not.

If users want to debug a configuration value, it might be better to implement a subcommand like git.
For example
```sh
$ navi config get shell.command
bash
```
If `navi config get` is useful, how about adding following `config` subcommand, instead of changing `config-path`?
```sh
navi config path # display current config path
navi config default-path # display default config path
navi config get `key.of.config` # display current value of specified key in configuration
navi config example # display example of configuration
```

The advantage of adding a `config` subcommand is that it is not a breaking chage.
In this case, it might be better to deprecate `navi info config-path`.

> Honestly, if you have a better method for debugging navi as a user (which can't edit the source code) I would greatly appreciate it.

The only debugging method I know of is to set `RUST_LOG` environment variable to `debug`.

---
By the way, you link this pull request to the issue #907 by comment `Fixes #907`, and the issue seems to suggest changing default config path or default cheats path.
Do you really intend to fix the issue #907 on this pull request?
If so, I think this is a breaking change, especially for users that use navi with default config path and default cheats path.

alexis-opolka:
I still find it weird to print a path that we don't use for a command that should output the path used but alright, two entrypoints might be better than only one.

In both cases, I think it's still a breaking change since the behaviour of the command changes.
(@denisidoro If we end up integrating this PR into the main branch, I think it's going to be necessary to clearly say in the release notes / changelog that the behaviour of `config-path` and `cheats-path` changed, otherwise it might be confusing for users)

Introducing a `config` subcommand might actually not be bad at all, I didn't think of that! But this will need its own PR (and issue to discuss what to expose). And we don't need to necessarily deprecate the `config-path` command as of right now but we would need a roadmap to predict when we deprecate the command and when we remove the support for it (and for some other deprecated features)

It seems to me that this part of navi needs to be more developed for solving issues related to its behaviour because as it stands now there is only the rust environment variable and I don't think it's either enough or well used to find the root cause of issues.
To me, we need a better debugging process for navi as I think it's not enough to easily solve issues and find where the issues lies.

---

For #907, I tried to fix it but over the iterations I think it still needs to be discussed, I'll check the code to see if there is any resulting aftereffects of my modifications but I'll be moving its fix into its dedicated PR after some discussions about the paths for the cheats and the configuration.

kit494way:
I expected `config-path` display current setting of config path,
but I think you are going to change `config-path` to display the source where the configuration was actually loaded from, right?

alexis-opolka:
@denisidoro Do you have any thoughts on this?

denisidoro:
I'd refrain from changing current behavior, to prevent breaking changes.

As for introducing new commands, I'm OK with whatever you folks think is the right design.

alexis-opolka:
Alright then, if you're also good with this @kit494way , I think that I'll migrate the current `info config-path` command into its own `config path`. (I will introduce a new issue and a new PR for the new `config` subcommands)

Then I'll revert the changes about `info config-path` but will introduce a deprecation warning which tends to the new `info default-config-path` which will make both commands outputting the same content allowing the users to migrate to the new command without this change being a breaking change. We will need to set a roadmap for the deprecated features.

kit494way:
@alexis-opolka I'm good with that.

alexis-opolka:
I'm closing down this PR in favor of #955 .

Issue #937, opened 2025-02-04
Refactor documentation + improvements


alexis-opolka:
**Is your feature request related to a problem? Please describe.**
The current documentation of Navi is not always up to date and is not well structured.

It might be the occasion to introduce two things:

- A Quickstart/Getting started guide for new users (important!)
- A static website (via github pages, for example) for the documentation

**Describe the solution you'd like**
Refactor the current documentation into a well structured `docs` folder and update its content.

**Describe alternatives you've considered**
Do not update the documentation and do not improve anything.

**Additional context**
As how issues are growing unresolved and features are added without being documented, it seems to me that we are not walking on the right path here.\
I would like for navi to be more documented and accessible to new users.

alexis-opolka:
@denisidoro I would like your thoughts on this.

denisidoro:
Sounds good. I'll gladly review related PRs if someone has the cycles to send them.

alexis-opolka:
@denisidoro I should have finished the new documentation by this weekend.
I'll surely send a PR this sunday.

alexis-opolka:
@denisidoro Just a quick question but how are defined Navi versions?
What is considered a major change? A minor change? A fix?
Do you have a limit of minor versions before a major version?

It would be great if you could explain it to me so I can include it in the documentation, thanks. 😄

denisidoro:
I don't think this is in set in stone but I'll share some examples.

> What is considered a major change?
Anything which introduces a major breaking change. The bash to rust rewrite was such an example. If I had to stop supporting current configs would also be a major change.

> Minor change
Almost everything.

> A fix
In general any release that I need to do right after another. If I released `x.y.0` but one day after I realized there was a typo in the help message and don't want to introduce any other features, I'd release `x.y.1`

alexis-opolka:
@denisidoro @kit494way It would be nice that we don't accept any more PRs regarding documentation after #961 until I've finished working on its refactoring since the structure and the content has changed quite a bit and it's going to be more and more difficult to merge the two types of structures.

It shouldn't take me long to finish, approximately two weeks at my current rate.

PR #941, opened 2025-02-09
Allow delimiter variable option to be set from config.yaml

### LINKED ISSUE #940: Allow the default delimiter to be changed in a configuration file
The first time I tried to use the `--column` option without the `--delimiter`, I was confused.
This is because I expected the default delimiter to be something like `\s+`, but it is acutally `\s\s+`.
https://github.com/denisidoro/navi/blob/5515367dc8a2a561d2d82b8352f729b1da4120d3/src/finder/post.rs#L46

Currently, when I use the `--column` option, I always specify the `--delimiter`, this is annoying.

Allow the default delimiter to be changed (e.g. `\s+`) in a configuration file.

Changing the default delimiter specified in source code is not preferable, as it would be a breaking change.

**Additional context**
Use cases are here.

```
% gh

# download release asset
gh release download <release_tag> -R <release_repo> -p <asset_pattern>

$ release_tag: gh release list -R <release_repo> --exclude-pre-releases --exclude-drafts --json tagName,publishedAt --jq '.[] | [.tagName, .publishedAt] | @tsv' --- --column 1 --delimiter '\t'
$ asset_pattern: gh release view -R <release_repo> --json assets --jq '.assets[].name'
```

```
% mise

# use local
mise use <tool>@<version>

$ tool: mise plugins ls -c --user -q
$ version: join -a 1 <(mise ls-remote <tool> | sort) <(mise ls -i <tool> | awk '{print $2 "\t*"}' | sort) | sort -k 1 -Vr --- --column 1 --delimiter "\s+"
```

kit494way:
Close #940 

After this change, if `delimiter_var` is specified in config.yaml, `parser::tests::test_parse_variable_line` test will fail.
https://github.com/denisidoro/navi/blob/5515367dc8a2a561d2d82b8352f729b1da4120d3/src/parser.rs#L349

### DIFF
--- a/docs/config_file_example.yaml
+++ b/docs/config_file_example.yaml
@@ -17,6 +17,7 @@ finder:
   command: fzf # equivalent to the --finder option
   # overrides: --tac # equivalent to the --fzf-overrides option
   # overrides_var: --tac # equivalent to the --fzf-overrides-var option
+  # delimiter_var: \s\s+ # equivalent to the --delimiter option that is used with --column option when you extract a column from the selected result for a variable
 
 # cheats:
 #   paths:
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -99,6 +99,10 @@ impl Config {
             .or_else(|| self.yaml.finder.overrides_var.clone())
     }
 
+    pub fn delimiter_var(&self) -> Option<String> {
+        self.yaml.finder.delimiter_var.clone()
+    }
+
     pub fn tealdeer(&self) -> bool {
         self.yaml.client.tealdeer
     }
--- a/src/config/yaml.rs
+++ b/src/config/yaml.rs
@@ -47,6 +47,7 @@ pub struct Finder {
     pub command: FinderChoice,
     pub overrides: Option<String>,
     pub overrides_var: Option<String>,
+    pub delimiter_var: Option<String>,
 }
 
 fn finder_deserialize<'de, D>(deserializer: D) -> Result<FinderChoice, D::Error>
@@ -158,6 +159,7 @@ impl Default for Finder {
             command: FinderChoice::Fzf,
             overrides: None,
             overrides_var: None,
+            delimiter_var: None,
         }
     }
 }
--- a/src/finder/structures.rs
+++ b/src/finder/structures.rs
@@ -66,6 +66,7 @@ impl Opts {
             overrides: CONFIG.fzf_overrides_var(),
             suggestion_type: SuggestionType::SingleRecommendation,
             prevent_select1: false,
+            delimiter: CONFIG.delimiter_var(),
             ..Default::default()
         }
     }


kit494way:
@alexis-opolka I created a pull request. Please update the documentation if necessary.

alexis-opolka:
I'll do that in a bit, thanks. ^^

alexis-opolka:
@kit494way Which takes precedence between using the configuration option and the `--delimiter` option in a cheat?

If someone defines its config file such as this:

```yaml
finder:
  command: fzf
  delimiter_var: \n
```

but for one variable uses:

```yaml
# This will pick the 3rd column and use the first line as header
docker rmi <image_id>

$ image_id: docker images --- --column 3 --header-lines 1 --delimiter '\s\s+'
```

which delimiter will be used?

---

On another note, I think that we might have to change the expected behaviour of the failing test.
I haven't yet checked the tests for Navi but I don't think that we should expect tests to fail after merging a new feature.
What are your thoughts on this @denisidoro?

---

On yet another note, I'm linking this PR with the issue for the refactoring of the documentation to not forget to update this feature in the new documentation. (https://github.com/alexis-opolka/navi/issues/1, #937)

alexis-opolka:
@kit494way I cannot push modifications to this PR as I'm not a maintainer in either this repository nor yours, you might be able to merge them into your repository.
The changes are in a separate branch on my repository: https://github.com/alexis-opolka/navi/tree/config-delimiter

kit494way:
@alexis-opolka 
> Which takes precedence between using the configuration option and the --delimiter option in a cheat?

The `--delimiter` option in a cheat has a higher precedence.
https://github.com/denisidoro/navi/blob/5515367dc8a2a561d2d82b8352f729b1da4120d3/src/parser.rs#L59

---
> I also think we should just call it delimiter inside the configuration file to keep a certain link with the --delimiter option.

I named the delimiter option `delimiter_var` based on the name of the overrides option.
There are two overrides options, `overrides_var` and `overrides`.
The `overrides_var` is used for variables in cheats and `overrides` is used for snippets.
The delimiter option added in this pull request is used for variables in cheats.

But I am not particular about whether the option name is `delimiter_var` or `delimiter`.
I don't mind changing to `delimiter` if there is a reason `delimiter` is better.

---
> I cannot push modifications to this PR as I'm not a maintainer in either this repository nor yours, you might be able to merge them into your repository.

Can you create a pull reaquest to `config-delimiter` branch on my repository?

denisidoro:
I'm confused - where are you using the field you just created?

kit494way:
@denisidoro 
The field is used here as default finder option for variables.
https://github.com/denisidoro/navi/blob/dc596526420539e44e0bd38db427b3b1a188db9e/src/finder/structures.rs#L69

The delimiter is used here to get a column from a selected result for a variable.
https://github.com/denisidoro/navi/blob/5515367dc8a2a561d2d82b8352f729b1da4120d3/src/finder/post.rs#L46

PR #944, opened 2025-02-22
Fix CI lint errors

kit494way:
Fix lint errors by `cargo clippy -- -D warnings`, and change ci.yml so that ci also runs on pull requests.
This prevent ci from failing on the master branch.


kit494way:
@denisidoro Would you review this?
It is useful to be able to notice that ci will fail before the changes are merged into the master.

alexis-opolka:
+1

It would be the start of a more complete CI workflow but knowing that changes will fail later on can greatly reduce the risks of merging PRs.

Issue #951, opened 2025-03-04
Multiple items in a cheatsheet cannot have the same description


gcarrillo:
**Describe the bug**
Before the refactor performed in https://github.com/denisidoro/navi/pull/760, it was possible to have multiple snippets in a cheatsheet that were different, but had the same description;  the various matching snippets would be displayed in the finder with identical descriptions.

After the refactor, only the first such snippet will be shown.

**To Reproduce**

1. Create a directory named "test", and inside of it, create a test cheatsheet containing multiple snippets with the same description:

```
$ cat test/foo.cheat
% tag_one, tag_two

# A non-unique description
snippet one

# A non-unique description
snippet two

# A non-unique description
snippet three
```

2.  Run Navi with the test cheatsheet:

`navi --path test
`

3.  Type "snippet" as input.

**Expected behavior**
Three snippets should match and be displayed in the finder.

**Actual behavior**
One snippet is displayed in the finder.

**Screenshots**

![Image](https://github.com/user-attachments/assets/95c0e30b-e086-492a-a729-ac91251faec2)

**Versions:**

- OS: Ubuntu 20.04
- Shell Version: Bash

**Additional context**
Note:  I have root caused this and will be submitting a fix.

gcarrillo:
In src/parser.rs:write_command(), we have:

```
let hash = item.hash();
if self.visited_lines.contains(&hash) {
    return Ok(());
}
self.visited_lines.insert(hash);
```
where item.hash() is defined as:
```
pub fn hash(&self) -> u64 {
    fnv(&format!("{}{}", &self.tags.trim(), &self.comment.trim()))
}
```

If two items hash the same, the second will not be written to the standard input of the finder.  Because the hash is computed only from an item's tags and comment, two different snippets with the same tags and comments are disallowed.  

We can fix this with the following hash function:
```
pub fn hash(&self) -> u64 {
    fnv(&format!("{}{}{}", &self.tags.trim(), &self.comment.trim(), &self.snippet.trim()))
}
```

PR #952, opened 2025-03-04
Allow cheatsheet items with same descriptions

gcarrillo:
Prior to the refactor performed in https://github.com/denisidoro/navi/pull/760, cheatsheet items could have the same tags and the same description, but a different snippet, and they would all be listed in the finder.

After the refactor, this behavior has changed and such items will now be treated as non-unique.

There appears to be no reason to require item descriptions to be distinct, so this patch restores the earlier behavior.

Fixes: #951

### DIFF
--- a/src/structures/item.rs
+++ b/src/structures/item.rs
@@ -18,6 +18,6 @@ impl Item {
     }
 
     pub fn hash(&self) -> u64 {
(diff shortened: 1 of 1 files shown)


alexis-opolka:
@denisidoro Was it intended to remove this feature? If not, I think you can merge this PR.

PR #943, opened 2025-02-17
Introduce a new documentation structure + Update its content

alexis-opolka:
Current status: WIP

---

This PR has just been made for @denisidoro (and others) to give me his thoughts on the changes I propose to make.

---

This PR is a part of the modification requested by #937

PR #955, opened 2025-03-11
Introduce the new default commands + Add a reference for the configuration + Add some debug information along

alexis-opolka:
This PR is the result of the discussions in #938.

It adds the following:

- Two new info commands ( `default-cheats-path` and `default-config-path` ) supposed to take over from the old info commands ( `cheats-path` and `config-path` )
- Introduces a deprecation notice on the old info commands
- Adds a reference inside the configuration on where it has been loaded from
- Adds some minor debug expressions
- Standardizes the way the functions retrieve without introducing a breaking change (normally)

### DIFF
--- a/src/commands/info.rs
+++ b/src/commands/info.rs
@@ -1,6 +1,6 @@
 use clap::Args;
 use clap::ValueEnum;
-
+use crossterm::style::Stylize;
 use crate::filesystem;
 use crate::prelude::*;
 
@@ -13,9 +13,13 @@ pub struct Input {
 #[derive(Debug, Clone, ValueEnum)]
 pub enum Info {
     CheatsExample,
+    ConfigExample,
+
     CheatsPath,
     ConfigPath,
-    ConfigExample,
+
+    DefaultCheatsPath,
+    DefaultConfigPath
 }
 
 impl Runnable for Input {
@@ -23,10 +27,19 @@ impl Runnable for Input {
         let info = &self.info;
 
         match info {
+            // Here should be the example commands
             Info::CheatsExample => println!("{}", include_str!("../../docs/cheat_example.cheat")),
-            Info::CheatsPath => println!("{}", &filesystem::default_cheat_pathbuf()?.to_string()),
-            Info::ConfigPath => println!("{}", &filesystem::default_config_pathbuf()?.to_string()),
             Info::ConfigExample => println!("{}", include_str!("../../docs/config_file_example.yaml")),
+
+            // Here should be the old deprecated default value commands
+            Info::CheatsPath => println!("{} Please use `info default-cheats-path` instead.\n\n\
+                {}","DEPRECATED:".red(), &filesystem::default_cheat_pathbuf()?.to_string()),
+            Info::ConfigPath => println!("{} Please use `info default-config-path` instead.\n\n\
+                {}", "DEPRECATED:".red(), &filesystem::default_config_pathbuf()?.to_string()),
+
+            // Here should be the default values (computed at compile time)
+            Info::DefaultCheatsPath => println!("{}", &filesystem::default_cheat_pathbuf()?.to_string()),
+            Info::DefaultConfigPath => println!("{}", &filesystem::default_config_pathbuf()?.to_string()),
         }
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


kit494way:
I think it is better to include `get_source` in the pull request of the code that will actually use it.

src/config/mod.rs
```diff
@@ -61,19 +62,48 @@ impl Config {
...
+                    debug!("ENV PATH: {}", self.env.path.as_ref().unwrap());
+                }
+
+                self.env.path.clone()
+            })
             .or_else(|| {
                 let p = self.yaml.cheats.paths.clone();
+
+                if !p.is_empty() {
+                    debug!("MULTIPLE YAML PATH: {}", p.as_slice().join(","));
```
kit494way, line 83: This can be moved to the else block following this.

src/config/yaml.rs
```diff
@@ -108,19 +109,36 @@ impl YamlConfig {
         serde_yaml::from_reader(reader).map_err(|e| e.into())
     }
 
-    pub fn get(env: &EnvConfig) -> Result<Self> {
+    pub fn get(env: &EnvConfig) -> Result<YamlConfig> {
         if let Some(yaml) = env.config_yaml.as_ref() {
-            return Self::from_str(yaml);
+            // We're getting the configuration from the environment variable `NAVI_CONFIG_YAML`
+            let mut cfg = Self::from_str(yaml)?;
+            cfg.source = format!("NAVI_CONFIG_YAML={}", yaml);
```
kit494way, line 116:
```suggestion
            cfg.source = format!("{}={}", env_var::CONFIG_YAML, yaml);
```
It is better to use `env_var::CONFIG_YAML` that is used for `env.config_yaml`.
This requires `use crate::env_var;`

src/config/yaml.rs
```diff
@@ -108,19 +109,36 @@ impl YamlConfig {
...
+
+            return Ok(cfg);
         }
         if let Some(path_str) = env.config_path.as_ref() {
+            // We're getting the configuration from a file given in the environment variable 'NAVI_CONFIG'
+
             let p = PathBuf::from(path_str);
-            return YamlConfig::from_path(&p);
+            let mut cfg = YamlConfig::from_path(&p)?;
+            cfg.source = format!("NAVI_CONFIG_YAML={:?}", p).to_string();
```
kit494way, line 125:
```suggestion
            cfg.source = format!("{}={}", env_var::CONFIG, p.to_string());
```
`NAVI_CONFIG_YAML` is typo of `NAVI_CONFIG` and it is better to use `env_var::CONFIG` that is used in `env.config_path`.
It is better to avoid using `{:?}` for anything other than debugging.
Unnecessary `to_string()`.

alexis-opolka:
@kit494way 

What I was trying to do was to get where it has been loaded from on a logical level and not on a technical level.
I want us to be able to deduct easily where issues are from regarding the configuration and that means being able to know at runtime what type of configuration is applied.
I don't think I should put the configuration value inside the source attribute since we can already debug it and we intend to introduce the config subcommands. Now it seems pretty useless to have a duplicata of values.

What I can propose is to remove the configuration values from `cfg.source` and only put the logical source, that will give us something like this:

| Logical origin | `cfg.source` value |
| :--------------: | :------------------: |
| default value (aka. built-in) | DEFAULT / BUILT-IN |
| default config file (YAML) | YAML_CONFIG_FILE |
| Environment Variable with YAML (NAVI_CONFIG_YAML) | ENV_NAVI_CONFIG_YAML |
| Environment Variable pointing to file (NAVI_CONFIG) | ENV_NAVI_CONFIG |

We can then get the real values since we know where to search.

This way also solves the issue I had on how to print the value.

src/config/mod.rs
alexis-opolka: Indeed, I'll do it tomorrow.

kit494way:
> What I can propose is to remove the configuration values from cfg.source and only put the logical source

I'm OK with that.

For the default config file, `cfg.source` value might be better as `DEFAULT_CONFIG_FILE` rather than `YAML_CONFIG_FILE`.

alexis-opolka:
Another review would be appreciated in case I missed something otherwise I think this PR should be good to be merged.

alexis-opolka:
I'll introduce the documentation on another PR, after #943 .

denisidoro (approved):
LGTM. Feel free to merge it.

Issue #969, opened 2025-04-02
Move required denisidoro/dotfiles scripts into navi


alexis-opolka:
We are moving the required scripts inside navi's repository.

PR #957, opened 2025-03-16
[SECURITY] Fixes malicious software reports and being flagged as trojan

alexis-opolka:
This PR aims to fix the security reports raised by Windows Defender and VirusTotal.

https://www.virustotal.com/gui/file/a4672b879ef32f9254c575dd548ea787bafe0827da1c6f93a65411a75a7121b2/detection

I am waiting to check a packaged version from an automated release on my fork but it seems the issue was from crossterm 0.28.0 which has been yanked ([see here](https://docs.rs/crossterm/0.28.0/crossterm/index.html)) from the creates registry in favor of 0.28.1([see here](https://docs.rs/crossterm/0.28.1/crossterm/index.html))

Fixes #869

Issue #869, opened 2023-12-19
Security concern about Windows binary (v2.23.0)


Geogboe:
**Describe the bug**
Want to first clarify that I believe this is a false positive but still wanted to bring it to the attention of the maintainers. When I tried to download the archive for this via Firefox I got a warning which prompted me to run it through virustotal which is flagging it for a multiple reasons -- all of which I don't fully understand not being familiar with the code. 

Here's the report: https://www.virustotal.com/gui/file/be1c45308c479db5d0ef6db49eefdbd41c2dbe543027807909e180b49fec2f0e

Edit: updated description

**To Reproduce**
NA 

**Expected behavior**
NA

**Screenshots**
NA

**Versions:**

- Version: 2.23.0
- OS: NA
- Shell Version: NA

**Additional context**

Here's what I tested

archive file:
Name: navi-v2.23.0-x86_64-pc-windows-gnu.zip
Size: 2611933 bytes (2550 KiB)
SHA256: 97539b0aa149c60dee1315d90e9339d84fb33ec80311b6d3c85aac07e5f22f22

executable:
Name: navi.exe
Size: 5172488 bytes (5051 KiB)
SHA256: be1c45308c479db5d0ef6db49eefdbd41c2dbe543027807909e180b49fec2f0e

VirusTotal report: https://www.virustotal.com/gui/file/be1c45308c479db5d0ef6db49eefdbd41c2dbe543027807909e180b49fec2f0e

I looked through the code base and build pipeline it looks like upx is being used to compress the binary: https://github.com/denisidoro/navi/actions/runs/7156992883/job/19487333651#step:5:208 and I'm thinking that might be making the bin more suspicious to scanners. I found a similar issue for another rust project: https://github.com/svenstaro/miniserve/issues/1210#issuecomment-1694579204 and even a pinned issue for upx itself: https://github.com/upx/upx/issues/437

These are pretty small binaries so I just wonder how much larger it would be without running it through upx and if that would remove some of the warnings?

denisidoro:
I can disable upx for Windows :)

Reyhn3:
Here's a report from [Hybrid Analysis](https://www.hybrid-analysis.com/sample/be1c45308c479db5d0ef6db49eefdbd41c2dbe543027807909e180b49fec2f0e) for version 2.23.0.

PR #971, opened 2025-04-07
Update the Continuous Deployment pipeline with a minimal amount of dependencies

alexis-opolka:
This PR tries to make the current pipeline of navi with the least amount of dependencies, which also include the `dotfiles` repository.

This PR is at a WIP state and is opened to receive feedback and required changes, although it will only be opened until 14/04/2025, after that it will be merged into master.
All platforms should be able to build without an issue now.

@denisidoro, if you feel like anything should be modified, let me know.

Fixes: #969
Closes: #957 
Fixes: #869

EDIT: This PR also closes #957 and its related issue #869


denisidoro (approved):
LGTM, but you dropped the call to `upx`. Could you please include it? (except for Windows: it causes some antivirus warnings, for some reason)

alexis-opolka:
Currently, the CI fails for aarch64_linux_android but it is a known issue from cross-rs/cross with rust versions greater than 1.68. (See https://github.com/cross-rs/cross/issues/1217)

alexis-opolka:
> LGTM, but you dropped the call to `upx`. Could you please include it? (except for Windows: it causes some antivirus warnings, for some reason)

Ok, since the size of the binaries weren't really big I didn't think of putting it back in the CI script but alright.

denisidoro:
> Currently, the CI fails for aarch64_linux_android but it is a known issue from cross-rs/cross with rust versions greater than 1.68. (See [cross-rs/cross#1217](https://github.com/cross-rs/cross/issues/1217))

For this to work, you need to use cross-rs's with a specific SHA. Here's what I did in another project of mine:

```sh
cargo install -f cross --git https://github.com/cross-rs/cross --rev ac4c11ce
```

alexis-opolka:
> > Currently, the CI fails for aarch64_linux_android but it is a known issue from cross-rs/cross with rust versions greater than 1.68. (See [cross-rs/cross#1217](https://github.com/cross-rs/cross/issues/1217))
> 
> For this to work, you need to use cross-rs's with a specific SHA. Here's what I did in another project of mine:
> 
> ```shell
> cargo install -f cross --git https://github.com/cross-rs/cross --rev ac4c11ce
> ```

It could be a solution but I don't think it's a healthy practice to have to unpin from a given tag and pin the project to a git ref.
It's better to wait a reasonable time for cross-rs/cross to achieve its 0.3 release and upgrade our CI then.
I already proposed my help to move forward with the 0.3 release we'll see how it goes.

If we're near the release date and cross-rs hasn't moved forward its 0.3 release we will temporarily pin our CI to a specifc git ref for the release but I would like this to be a one-time workaround.

alexis-opolka:
@denisidoro I'm going to add back UPX but until I've found how you installed it and used it I'm merging this PR in order to close the related issues since they are not dependent of the use of UPX.

PR #977, opened 2025-04-24
feat: Variables history

soar:
# what

If env variable `NAVI_VARIABLES_HISTORY_FILE` is set, stores information about entered variables, so they will be suggested next time the snippet is selected.

This is PoC, but in the future it can be expanded:
- sort suggestions according to the last input time
- option to enable/disable it per variable
- and much more

# demo

[asciicast](https://asciinema.org/a/716826)

navi-variables

### DIFF
--- a/src/env_var.rs
+++ b/src/env_var.rs
@@ -19,7 +19,7 @@ pub const FINDER: &str = "NAVI_FINDER";
 pub const CONFIG: &str = "NAVI_CONFIG";
 pub const CONFIG_YAML: &str = "NAVI_CONFIG_YAML";
 
-pub const VARIABLES_HISTORY_FILE: &str = "NAVI_VARIABLES_HISTORY_FILE";
+pub const VAR_HISTORY: &str = "NAVI_VAR_HISTORY";
 
 pub fn parse<T: FromStr>(varname: &str) -> Option<T> {
     if let Ok(x) = env::var(varname) {
--- a/src/env_var.rs
+++ b/src/env_var.rs
@@ -19,6 +19,8 @@ pub const FINDER: &str = "NAVI_FINDER";
 pub const CONFIG: &str = "NAVI_CONFIG";
 pub const CONFIG_YAML: &str = "NAVI_CONFIG_YAML";
 
+pub const VARIABLES_HISTORY_FILE: &str = "NAVI_VARIABLES_HISTORY_FILE";
+
 pub fn parse<T: FromStr>(varname: &str) -> Option<T> {
     if let Ok(x) = env::var(varname) {
         x.parse::<T>().ok()
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -2,12 +2,14 @@ pub use crate::common::fs::{create_dir, exe_string, read_lines, remove_dir};
 use crate::env_var;
 use crate::parser::Parser;
 use crate::prelude::*;
+use std::io::Write;
 
 use crate::structures::fetcher;
 use etcetera::BaseStrategy;
 use regex::Regex;
 
 use std::cell::RefCell;
+use std::env;
 use std::path::MAIN_SEPARATOR;
 
 use walkdir::WalkDir;
@@ -85,6 +87,53 @@ pub fn cheat_paths(path: Option<String>) -> Result<String> {
         Ok(p)
     } else {
         Ok(default_cheat_pathbuf()?.to_string())
+    }
+}
+
+pub fn variables_history_pathbuf() -> Option<PathBuf> {
+    if let Ok(v) = env::var(env_var::VARIABLES_HISTORY_FILE) {
+        let pathbuf = PathBuf::from(v);
+
+        if !pathbuf.exists() {
+            File::create(&pathbuf).unwrap_or_else(|_| panic!("Unable to create file: {}", pathbuf.display()));
+        }
+
+        Some(pathbuf)
+    } else {
+        None
+    }
+}
+
+pub fn get_variable_history(variable: &str) -> Vec<String> {
+    if let Some(pathbuf) = variables_history_pathbuf() {
+        let file = std::fs::File::open(pathbuf).expect("Unable to open history file");
+        let reader = std::io::BufReader::new(file);
+
+        reader
+            .lines()
+            .filter_map(|line| line.ok())
+            .filter_map(|line| {
+                let parts: Vec<&str> = line.splitn(2, ':').collect();
+                if parts.len() == 2 && parts[0] == variable {
+                    Some(parts[1].to_string())
+                } else {
+                    None
+                }
+            })
+            .collect()
+    } else {
+        vec![]
+    }
+}
+
+pub fn save_variable_history(variable: &str, value: &str) {
+    if let Some(pathbuf) = variables_history_pathbuf() {
+        let mut file = std::fs::OpenOptions::new()
+            .append(true)
+            .create(true)
+            .open(pathbuf)
+            .expect("Unable to open history file");
+        writeln!(file, "{}:{}", variable, value).expect("Unable to write to history file");
     }
 }
 
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -91,7 +91,7 @@ pub fn cheat_paths(path: Option<String>) -> Result<String> {
 }
 
 pub fn variables_history_pathbuf() -> Option<PathBuf> {
-    if let Ok(v) = env::var(env_var::VARIABLES_HISTORY_FILE) {
+    if let Ok(v) = env::var(env_var::VAR_HISTORY) {
         let pathbuf = PathBuf::from(v);
 
         if !pathbuf.exists() {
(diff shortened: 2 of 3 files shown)


alexis-opolka:
Hi @soar ,
I haven't taken a look at your code yet but I have some issues regarding the idea. This being opt-in is also a good thing.

I find it great that we could remember the values typed but It shouldn't be enabled only by setting an environment variable.
I would prefer to have an entry inside the configuration file and maybe an environment variable but not the other way.
Regarding the variable name, I would go with `NAVI_VAR_HISTORY` instead of `NAVI_VARIABLES_HISTORY_FILE`.
It's shorter and we can just write into a file (`navi-var.history`) in the directory next to the configuration file, like how the log/debug file is handled.

This feature, if accepted, will need to be documented, so it will have to wait for #943 to be merged.

I think we should make it visible that it's from the history and not a pre-defined value list, we need something to clearly differentiate both.
Also, how do you take into account the different types of variables?
You can have empty variables, pre-defined variables with computed values and have the possibility to have multiple cheats using the same variable name but in different contexts which can require completely different types of values.

Because variables with pre-defined values won't require to have an history of entered values.

Does it impact in some way the scripting capabilities? Have you tested it?

src/env_var.rs
```diff
@@ -19,6 +19,8 @@ pub const FINDER: &str = "NAVI_FINDER";
 pub const CONFIG: &str = "NAVI_CONFIG";
 pub const CONFIG_YAML: &str = "NAVI_CONFIG_YAML";
 
+pub const VARIABLES_HISTORY_FILE: &str = "NAVI_VARIABLES_HISTORY_FILE";
```
alexis-opolka, line 22: As said before, the name would need to change.

src/filesystem.rs
```diff
@@ -88,6 +90,53 @@ pub fn cheat_paths(path: Option<String>) -> Result<String> {
     }
 }
 
+pub fn variables_history_pathbuf() -> Option<PathBuf> {
+    if let Ok(v) = env::var(env_var::VARIABLES_HISTORY_FILE) {
```
alexis-opolka, line 94: I think we should use option_env! instead of env::var and use the environment variable to take precedence on the configuration value but we should always fallback to the configuration value.

This means that you need to make an entry inside the configuration.

src/filesystem.rs
```diff
@@ -88,6 +90,53 @@ pub fn cheat_paths(path: Option<String>) -> Result<String> {
...
+        let pathbuf = PathBuf::from(v);
+
+        if !pathbuf.exists() {
+            File::create(&pathbuf).unwrap_or_else(|_| panic!("Unable to create file: {}", pathbuf.display()));
```
alexis-opolka, line 98: I'm not sure about the use of the panic! macro here.
We can just not save the entry to the filesystem and output an error message afterward.
Navi doesn't rely on the variables history to work.

There is also the fact that we create the file inside the function that should just send back the path.
We should migrate this part to the save function.

src/filesystem.rs
```diff
@@ -88,6 +90,53 @@ pub fn cheat_paths(path: Option<String>) -> Result<String> {
     }
 }
 
+pub fn variables_history_pathbuf() -> Option<PathBuf> {
```
alexis-opolka, line 93: I would prefer the function to have the `default` prefix because the logic is the same as the other default functions.

alexis-opolka (changes requested):
The overall looks promising although there are certain points that should be modified in my opinion.

alexis-opolka:
The `default` prefix does not mean to be enabled by default fyi, it only means that it is a value of reference. A default value given by the logic of a function.

It is used for the cheats path which has a default path but can be modified by the users to point to another path.

`default_` != Enabled by default.

I hope this solves your confusion on this subject.

alexis-opolka:
> @alexis-opolka I still feel a misunderstanding here. Do we both agree, that if I change `env::var` to `option_env!` as you are requesting, users won't be able to disable this or change the file path? My point is to make it optional. Your suggestions lead us to forcing users to use this feature, which I want to avoid for now.

No @soar, what I mean by using `option_env!` is to enable this feature only if navi is compiled with the variable set.

See this function for the default cheats path:

https://github.com/denisidoro/navi/blob/138b7e0b0086f6c2d3cfd8274d7e616c5e980ab8/src/filesystem.rs#L55-L67

We are using `option_env!` to overwrite the default value if the environment variable was set during compilation.

So take for example `NAVI_PATH` being set to `/my/path`:

- `NAVI_PATH` set during compilation => cheats path is set to `/my/path`
- `NAVI_PATH` not set during compilation => cheats path is set to `~/.local/share/navi/cheats/` on linux (= default behaviour)

I want this feature to work in the same way.

If `NAVI_VAR_HISTORY` is not set during compilation => we **DO NOT** save any variable input
If `NAVI_VAR_HISTORY` si set during compilation => we **DO** save any variable input

I might repeat myself but **WE DO NOT** enable this feature by default.

The logic is the same as the one you currently have except we don't rely on an environment variable while running but during the compilation.

alexis-opolka:
@soar 

Maybe some code will help you understand what I mean:

```rust
pub fn default_variables_history_pathbuf() -> Result<PathBuf> {
    let mut pathbuf = get_data_dir_by_platform()?;

    pathbuf.push("navi");
    pathbuf.push("var.history");

    Ok(pathbuf)
}

pub fn get_variable_history(variable: &str) -> Vec<String> {
    // If the feature isn't enabled, exit immediately
    if ! is_var_history_enabled() {
        return vec![]
    }

    let pathbuf = default_variables_history_pathbuf().unwrap();
    let file  = match File::open(pathbuf) {
        Ok(file) => file,
        Err(_) => return vec![],
    };

    let reader = BufReader::new(file);

    reader
        .lines()
        .map_while(|line| line.ok())
        .filter_map(|line| {
            let parts: Vec<&str> = line.splitn(2, ':').collect();
            if parts.len() == 2 && parts[0] == variable {
                Some(parts[1].to_string())
            } else {
                None
            }
        })
        .collect()
}

pub fn save_variable_history(variable: &str, value: &str) {
    // If the feature isn't enabled, exit immediatly
    if ! is_var_history_enabled() {
        return;
    }

    let pathbuf = default_variables_history_pathbuf().unwrap();

    // Create the file if it doesn't exist
    if !pathbuf.exists() {
        File::create(&pathbuf).unwrap();
    }

    let mut file = std::fs::OpenOptions::new()
        .append(true)
        .create(true)
        .open(pathbuf)
        .expect("Unable to open history file");
    writeln!(file, "{}:{}", variable, value).expect("Unable to write to history file");
}
```

and

```rust
pub fn is_var_history_enabled() -> bool {
    if let Some(history) = option_env!("NAVI_VAR_HISTORY") {
        if history == "true" { 
            return true
        }
    }

    false
}
```

This way, the feature is enabled only if we set the variable during the compilation of navi.
That means it will be disabled by default.

alexis-opolka:
Hi @soar ,

I think this is indeed the point of misunderstanding, I might not have explained well and I'm sorry for that.
What I wanted to say at the start is that I do not want a feature in navi to not be either registered in the Config struct or with setting an environment variable during compilation. Let me explain what I mean by that.

In the line you gave me, the EnvConfig struct is using a userspace variable during the execution but those entries in the configuration are not only present in EnvConfig but in YamlConfig, Config and maybe in ClapConfig.

For an experimental feature, I do not want to rely on only one way to enable it and surely not a userspace environment variable without having an entry whatsoever in one of the Config structs. I need to have a defined way to know if the feature is enabled or not, be it for debugging purposes, more experimentation, whatsoever the goal afterwards.

This is the main reason of why I asked you to put an entry in the configuration and being in the configuration does not mean being enabled by default.

By having an entry in the configuration, say in `YamlConfig`, `EnvConfig` and `Config`, you'd have two different ways to enable it without having to compile navi.
`YamlConfig` can be set with an environment variable or with the yaml configuration file in `~/.config/navi/config.yaml` on linux, `EnvConfig` handles all the environment variables navi supports and `Config` is the central piece of this puzzle and handles the default values if no values exist in any of the above structs.

That is why I proposed you, if you wanted to enable the feature with an environment variable and only that to instead use the environment variable during compilation.

Once again, I understand the risks tied to this feature in its current state and I do not want this feature to be enabled by default, I want this feature to be disabled as a default just like you.

I hope this solves a part or all of the confusion between us, if this is not the case do not hesitate to ask me more questions.

alexis-opolka:
I'm running the CI in the meanwhile since I already reviewed your code.

alexis-opolka:
@soar If you need me to show you via a code snippet what I mean don't hesitate to ask! 😄

alexis-opolka:
Hi @soar,
Any news on this PR?

alexis-opolka:
Hi, I would still like to keep your code for later since I agree that this is a great feature.
I'll be closing this PR and will make a new branch called `feat/variable-history`, can you open another one then and target the new branch?

That way we will have a trace left for future work on that feature.

PR #980, opened 2025-05-07
Update documentation links

ipwnponies:
Updating broken links from refactoring in https://github.com/denisidoro/navi/pull/943

I tested these by verifying each link resolves correctly, on my fork.


alexis-opolka:
Hi @ipwnponies ,

Great catch and I thank you! ^-^
I made a small modification for the links to point to the files instead of the directories since in an offline environment pointing on the directories will not have the desired effect.

alexis-opolka (approved):

PR #998, opened 2025-11-26
fix: bash widget sanitization for &

grota:
In the bash widget readline code make sure that bash's `patsub_replacement` is temporarily unset to avoid having `${input}_NAVIEND` in the final command when the expanded $replacement contains `&`


alexis-opolka:
Oh, just to let you know, I plan to add a test for your use case to ensure it doesn't break in the future.

Issue #964, opened 2025-03-25
Improve Windows default behaviour


alexis-opolka:
A lot of issues are opened for windows regarding the default behavior of navi on this platform.
Mainly regarding that navi tries to work with bash when it is not the default to have a shell such as Bash on a windows install.

We are also proposing to install a default cheatsheet which is better suited for linux environments.\
It might also be a good idea to introduce a specific cheatsheet for windows, disable the option to download it or show a warning that it is not meant to be downloaded on this platform.

See: #897 #794 #880 #885 #897 #926

PR #1007, opened 2026-02-17
Better Windows support

Wind010:
Better Windows support:
https://github.com/denisidoro/navi/issues/964

Default to use `PowerShell.exe` without configuration of `config.yaml`.  
Added additional configurations for Shell.
Support `clip.exe` when used on Windows.

### DIFF
--- a/src/common/clipboard.rs
+++ b/src/common/clipboard.rs
@@ -17,20 +17,22 @@ pub fn copy(text: String) -> Result<()> {
       Ok(())
    } else {
       // Use bash/zsh/fish/etc logic
-      let mut script = String::new();
-      script.push_str("exst() { type \"$1\" &>/dev/null; }\n");
-      script.push_str("_copy() {\n");
-      script.push_str("  if exst pbcopy; then pbcopy\n");
-      script.push_str("  elif exst xclip; then xclip -selection clipboard\n");
-      script.push_str("  elif exst clip.exe; then clip.exe\n");  // # Could be used on WSL or GitBash
-      script.push_str("  else exit 55; fi\n");
-      script.push_str("}\n");
-      script.push_str(&format!(
-         "read -r -d '' x <<\"{EOF}\"\n{text}\n{EOF}\n",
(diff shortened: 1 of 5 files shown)


src/common/clipboard.rs
```diff
@@ -2,38 +2,40 @@ use crate::common::shell::{self, ShellSpawnError, EOF};
...
+      let mut cmd = std::process::Command::new("cmd.exe");
+      cmd.arg("/C").arg("clip.exe");
+      let mut child = cmd.stdin(std::process::Stdio::piped()).spawn()
+         .map_err(|e| ShellSpawnError::new("clip.exe", e))?;
+      if let Some(stdin) = child.stdin.as_mut() {
+         use std::io::Write;
+         stdin.write_all(text.as_bytes())?;
+      }
+      child.wait()?;
+      Ok(())
```
alexis-opolka: While I understand that using clip.exe via cmd.exe is easy and surely a mature/robust way of doing it, I can't help to think that if we go forward to having a stable way of using navi under windows systems, we can do it in a better way.

First, as PowerShell is becoming more and more robust as time passes, why stop ourselves at using CMD?
We can always fall back to using CMD in case something goes wrong with the PowerShell API.

Second, is it the best way to implement this? I haven't dabbled in this part of the codebase that much but it might be better to use an API instead of the shell command directly, no?

src/config/yaml.rs
```diff
@@ -77,6 +77,8 @@ pub struct Search {
 pub struct Shell {
     pub command: String,
     pub finder_command: Option<String>,
+    pub show_command: Option<bool>,
+    pub command_print_color: Option<Color>,
```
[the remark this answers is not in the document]
Wind010: Will remove and add a separate feature ticket.

src/common/clipboard.rs
alexis-opolka: Yeah, if you could make a FR for that so that it can be done for a future version, that would be awesome. Thanks ^^

Issue #1009, opened 2026-02-25
Print the selected command to standard out as a configurable option in config.yaml


Wind010:
**Is your feature request related to a problem? Please describe.**
It would be nice to see the selected command printed for posterity and selection in the terminal.

**Describe the solution you'd like**
Print the command to standard out as a configurable option in config.yaml.  Color of that command can be customized so it can stand out more.

**Describe alternatives you've considered**
Just CTRL+Y to copy the command.

**Additional context**
Would like to add this feature

PR #1011, opened 2026-03-07
feat: add --print-command and --print-cheat flags

kavau:
## Summary

This PR adds two new optional flags that print information about the selected
cheat to stdout before executing it.

### `--print-command` / `-P`

Prints the resolved command before executing it:

```
% navi --print-command -q "apt list"
navi> apt list --upgradeable
Listing... Done
code/stable 1.110.1-1772839555 arm64 [upgradable from: ...]
```

Useful for keeping a visible record of what was run in a terminal session.

### `--print-cheat` / `-C`

Prints a formatted block with the comment and command before executing it:

```
% navi --print-cheat -q "apt list"
──────────────────────────────────────
# Check updates for installed packages
apt list --upgradeable
──────────────────────────────────────
Listing... Done
code/stable 1.110.1-1772839555 arm64 [upgradable from: ...]
```

The separator width adapts to the longest line (capped at 80 chars). The
comment and command are coloured using `crossterm` (already a dependency).

## Motivation

Selecting a cheat, having navi execute it, and then seeing only the command output but not the command itself, feels like "flying blind". It's nicer and safer to have a record of exactly what happened in the terminal output.

## Changes

- `src/config/cli.rs` — new `--print-command` (`-P`) and `--print-cheat` (`-C`) flags, plus examples in `--help`
- `src/config/mod.rs` — `print_command()` and `print_cheat()` accessors on `Config`
- `src/commands/core/actor.rs` — print logic before `spawn()`
- `docs/usage/README.md` — new sections documenting `--print-command` and `--print-cheat` 

## Notes

- `rust-toolchain.toml` was bumped from `1.81.0` to `1.88.0` to resolve a
  build failure caused by transitive dependencies requiring newer editions
  (`getrandom 0.4`, `home 0.5.12`). This seems like a pre-existing issue
  worth fixing regardless of this PR.
- When both `-C` and `-P` are passed, `-C` takes precedence.

MAINTAINER EDIT: Partially answers to #1009

### DIFF
--- a/src/config/cli.rs
+++ b/src/config/cli.rs
@@ -53,6 +53,10 @@ pub(super) struct ClapConfig {
     #[arg(short = 'P', long)]
     pub print_command: bool,
 
+    /// Prints the full cheat (comment + command) before executing it
+    #[arg(short = 'C', long)]
+    pub print_cheat: bool,
+
     /// Returns the best match
     #[arg(long)]
     pub best_match: bool,
--- a/docs/usage/README.md
+++ b/docs/usage/README.md
@@ -22,6 +22,27 @@ navi --print-command
 
 This is useful to keep a visible record of what was run, e.g. in a terminal session log.
 
+## Printing the full cheat before executing
+
+You can print a formatted block showing the comment and resolved command before execution using `--print-cheat` (or `-C`):
+
+```sh
+navi --print-cheat
+```
+
+Example output:
+
+```
+──────────────────────────────────────
+# Check updates for installed packages
+apt list --upgradeable
+──────────────────────────────────────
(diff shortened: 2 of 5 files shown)


src/config/cli.rs
```diff
@@ -48,6 +49,14 @@ pub(super) struct ClapConfig {
     #[cfg(not(feature = "disable-command-execution"))]
     pub print: bool,
 
+    /// Prints the command to stdout before executing it
+    #[arg(short = 'P', long)]
```
denisidoro: I'd rather not add the `short` option, because we'll need to maintain it forever (unless there's a breaking change).

PR #1027, opened 2026-05-18
Fix CI test reliability and add tmux shell plugin tests (#1014)

GauravS11112003:
## Summary

This PR adds a **dedicated tmux-driven test suite** for the **bash**, **zsh**, and **fish** shell widgets under [`shell/`](https://github.com/denisidoro/navi/tree/master/shell), so plugin behaviour is exercised inside real interactive shells instead of only checking that `navi widget <shell>` prints something containing `navi`.

It also **hardens CI** around third-party `tldr` usage and **non-interactive** test runs (no TTY), so the main [`./tests/run`](https://github.com/denisidoro/navi/blob/master/tests/run) job stays reliable on GitHub Actions.

## What’s included

- **[`tests/shell/run`](https://github.com/denisidoro/navi/blob/master/tests/shell/run)** — drives each shell in tmux, sources the matching plugin, sends **Ctrl+G**, and asserts on **sentinel output** from tiny fixtures in [`tests/shell/cheats/`](https://github.com/denisidoro/navi/tree/master/tests/shell/cheats).
- **[`tests/shell/lib.bash`](https://github.com/denisidoro/navi/blob/master/tests/shell/lib.bash)** — tmux helpers (session start/stop, `send-keys`, pane capture, wait-for regex).
- **Minimal rc files** under [`tests/shell/rc/`](https://github.com/denisidoro/navi/tree/master/tests/shell/rc) so plugins load in a predictable interactive session.
- **[`.github/workflows/shell-tests.yml`](https://github.com/denisidoro/navi/blob/master/.github/workflows/shell-tests.yml)** — **path-filtered** workflow (only when `shell/**`, `tests/shell/**`, or that workflow file changes) so unrelated PRs do not pay for zsh/fish/tmux setup.
- **Regression-style check** for **multi-line snippet handling** (same class of bug as discussed in [#1010](https://github.com/denisidoro/navi/pull/1010)): two-line fixtures must still produce **two** distinct sentinel lines after the widget runs.

## CI / integration test notes

- **`tldr`**: CI installs **tealdeer** and sets [`client.tealdeer: true`](https://github.com/denisidoro/navi/blob/master/tests/config.yaml) for the test config; the cache is primed with **`tldr --update`** before the third-party test.
- **`stty`**: Integration wrapper only runs `stty sane` when **stdin is a TTY**, so headless CI does not open `/dev/tty` and fail under `set -e`.
[...]

PR #1032, opened 2026-07-13
fix: preserve multiline snippets in fish by clearing IFS locally

leoraclet:
## Problem

When using the latest up-to-date widget in Fish shell, multi-line input is rendered as a single line instead of preserving line breaks.

## Fix

Declare a local empty IFS before capturing the output of `navi`.

Without this, Fish splits multiline command substitutions into a list, causing `commandline --replace` to flatten the snippet into a single line.

Using a local empty IFS preserves the multiline output while keeping the change scoped to `_navi_smart_replace`.

### DIFF
--- a/shell/navi.plugin.fish
+++ b/shell/navi.plugin.fish
@@ -1,6 +1,7 @@
 function _navi_smart_replace
     set --local query (commandline --current-process | string trim)
     set --local version_parts ""
+    set --local IFS
     if test -n "$version"
         set version_parts (string split '.' $version)
     else


jackandroselv:
您好！您的邮件我已收到，我会尽快的查阅邮件。谢谢您的来信！

alexis-opolka:
Hi @leoraclet !

I'll check it out tomorrow and leave you a review for sure, it will surely be merged after #1027 though. ^-^

alexis-opolka (approved):
It's good, thanks!

PR #1034, opened 2026-07-20
Fix variable query override precedence

### LINKED ISSUE #920: --query not working as intended
**Describe the bug**
I cannot make --query work. I can use --fzf-overrides to set the query there, but I can't use variables inside the query in that case.

**To Reproduce**

1. Add variable `$ title:  --- --query "test"`
2. Run the cheat that uses that variable
3. The word "test" is not the initial query

**Expected behavior**
The --query argument should set the initial query of a variable

**Versions:**

- OS: Manjaro
- Shell Version: zsh 5.9 (x86_64-pc-linux-gnu)

zhangli091011:
## Summary

- preserve a cheatsheet variable query when no `<variable>__query` environment override exists
- keep the environment query as the higher-precedence override
- test both precedence paths and variable-line query parsing

## Testing

- `cargo fmt --all -- --check`
- `cargo test` (Rust 1.81 Linux container)
- `cargo check --all-targets`
- `cargo clippy -- -D warnings` (Windows and Rust 1.81 Linux container)

Fixes #920

## Disclosure

This pull request was implemented and tested with assistance from OpenAI's OpenCode coding agent.

### DIFF
--- a/src/commands/core/actor.rs
+++ b/src/commands/core/actor.rs
@@ -12,6 +12,10 @@ use crate::structures::item::Item;
 use shell::EOF;
 use std::process::Stdio;
 
+fn query_with_override(query: Option<String>, query_override: Option<String>) -> Option<String> {
+    query_override.or(query)
+}
+
 fn prompt_finder(
     variable_name: &str,
     suggestion: Option<&Suggestion>,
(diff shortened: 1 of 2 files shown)


alexis-opolka:
Hi @zhangli091011 !

Thanks for the PR, I'll check it out in the weekend, and I appreciate the disclosure about the use of AI!

alexis-opolka (approved):

PR #1039, opened 2026-08-13
fish: stop the widget reopening fzf over its own best match

### LINKED ISSUE #1017: Navi widget for fish
So I saw a closed issue regarding navi widget for fish. The readme checklist shows that its not there for fish even though navi widget fish | source is there. I tried it and its not properly working as shown in the demo video. Wanted to get more clarification as this tool looks interesting with the widget and would like to try it out.

GauravS11112003:
`Ctrl+G` in fish silently replaced the prompt with a `--best-match` result and then immediately reopened fzf on top of it, so the widget never behaved the way the demo shows. This fixes that, restores prefix-preserving replacement, and adds regression coverage.

Fixes #1017

## Root cause 1: a variable fish had already destroyed

```fish
if test -n "$query"
    set --local best_match (navi --print --query "$query" --best-match)
    ...
end

if test -z "$best_match"      # always true
```

fish erases `--local` variables when their enclosing block ends, and `if` is a block, so `$best_match` was always empty by the time the fallback tested it. Every `Ctrl+G` on a non-empty prompt ran navi twice: splice in the best match, then open the picker over what it had just inserted. This came in with aa87ad4 (#1032), which swapped an early `return` for a check on the now out-of-scope variable.

The fix hoists the result to function scope, so the fallback only runs when `--best-match` genuinely found nothing.

## Root cause 2: the whole buffer was being replaced

#1010 changed the splice to `commandline --replace`, which rewrites the entire command line rather than the command under the cursor:

```
before:   true | ping   ->   echo "..."
after:    true | ping   ->   true | echo "..."
```

The flattening that #1010 set out to fix was really `$best_match` being an unquoted fish list. `string collect` addresses that at the source, so the splice can go back to `commandline --current-process` and keep the left-hand side of the pipe. That also let me drop the empty `$IFS`, which silently changes `read` semantics for anything else in scope.

## A second press still reaches the picker

With the fallthrough fixed, a non-empty prompt would otherwise only ever get `--best-match`. zsh solves this with `$LASTWIDGET`; fish now remembers the buffer it wrote, so pressing `Ctrl+G` again on a result you haven't edited opens the picker seeded with your original query.

## Smaller things

- `$FISH_VERSION` replaces `$version` plus an unreachable `fish --version` fallback, and the `-ge 4` comparison no longer errors on a non-numeric major
- `bind` is guarded by `status is-interactive`
- Fish and Zsh get their checkmark in the widget support table. That was the other half of #1017: the table said unsupported while the same page documented how to install them. Added a short section on the two-press behaviour too.

## Tests

Three new cases in `tests/shell/run`, already covered by the existing `shell/**` path filter in CI. All three fail against the old plugin with exactly the symptoms above and pass against the new one. The pre-existing `query_driven` and `multiline_snippet` cases passed either way, which is how #1017 slipped through.

- `best_match_skips_picker (#1017)` (zsh, fish): the snippet lands in the buffer and fzf never appears
- `repeat_press_opens_picker (#1017)` (zsh, fish): a second `Ctrl+G` opens the picker and the buffer survives accepting it
- `pipe_prefix (#1017)` (bash, zsh, fish): the left-hand side of a pipe survives the replacement and the result runs

`./tests/shell/run` passes 16/16 locally against fish 4.8.1, zsh 5.9, bash 5.3, tmux 3.7b and fzf 0.74.2. CI covers fish 3.7 on Ubuntu.

Also checked by hand: empty prompt opens the picker, a typed query gets its best match, a second press opens the picker, the pipe prefix survives, multi-line snippets stay on separate lines, and vi mode keeps its `[I]` indicator after the repaint.

### DIFF
--- a/docs/widgets/README.md
+++ b/docs/widgets/README.md
@@ -10,10 +10,18 @@ Widgets are 3rd-party contributions and integrate Navi with 3rd-party software s
 | Shell      | Navi support       |
 |------------|--------------------|
 | Bash       | :white_check_mark: |
-| Fish       |                    |
-| Zsh        |                    |
+| Fish       | :white_check_mark: |
+| Zsh        | :white_check_mark: |
 | NuShell    | :white_check_mark: |
 | PowerShell | :white_check_mark: |
+
+## Fish Widget
+
+The fish widget replaces the command under the cursor, so anything to the left of a pipe or a `&&` is left alone.
+
+- on an empty prompt, `Ctrl+G` opens the picker
+- with something typed, `Ctrl+G` replaces it with the best match for it, without opening the picker
+- pressing `Ctrl+G` again, before editing that result, opens the picker with your original text as the query
 
 ## PowerShell Widget
 
--- a/shell/navi.plugin.fish
+++ b/shell/navi.plugin.fish
@@ -1,13 +1,21 @@
 function _navi_smart_replace
-    set --local query (commandline --current-process | string trim)
-    set --local version_parts ""
-    set --local IFS
-    if test -n "$version"
-        set version_parts (string split '.' $version)
-    else
-        set version_parts (string split '.' (string match -r '\d+\.\d+\.\d+' (fish --version)))
+    set --local buffer (commandline --current-buffer | string collect)
+    set --local process (commandline --current-process | string collect)
+    # --current-process hands back the whitespace separating the command from a
+    # preceding pipe or operator, so keep it to splice it back in afterwards
+    set --local indent (string match --regex '^\s*' -- "$process")
+    set --local query (string trim -- "$process")
+
+    # A second press on a result we spliced and the user hasn't edited means the
+    # best match was wrong, so fall through to the picker seeded with the query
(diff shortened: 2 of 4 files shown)


GauravS11112003:
@denisidoro @alexis-opolka just a followup on the PR. lmk if you need any changes. PS. used cursor for this one, looks fine to me but would appreciate an extra eye.

alexis-opolka:
Hi, I'll check it out today.

Sorry for the delay, I'm pretty busy this summer.

GauravS11112003:
No problem!

GauravS11112003:
Just a followup. @alexis-opolka @denisidoro

alexis-opolka:
Yes @GauravS11112003 thanks so much for the ping, I've installed fish in order to test it out before being able to do a correct review probably after work today. Do ping me again if I fail to do it today. ^-^

Sorry again for the long time you have and had to wait for the reviews.

(I think it would be better to not ping Denis on the PRs / issues regarding the shell widgets. ^^)

PR #1040, opened 2026-08-15
fix(tldr): name the config requirement and the moved tealdeer repo

### LINKED ISSUE #1036: Clarify VERSION_DISCLAIMER in `src/clients/tldr.rs`
Context: When using `navi --tldr <cmd>`.

The error message in this snippet needs clarification:
https://github.com/denisidoro/navi/blob/f7330b9ad5bd95b7d1a3c96d00e0a77deb589147/src/clients/tldr.rs#L10-L12

The message suggests using TLDR `tealdeer`, but it doesn't mention that the `client.tealdeer: true` configuration is a requirement for it to work (see https://github.com/denisidoro/navi/pull/873 and https://github.com/denisidoro/navi/issues/902#issuecomment-2263011184).

Without this requirement, the message feels confusing.
Even after adding `tealdeer` to the system, `navi --tldr <cmd>` still doesn't work and displays the same message.

---

I found `tealdeer` at a different GitHub repository address: https://github.com/tealdeer-rs/tealdeer

It looks like `tealdeer` has migrated to a dedicated repository. This could be updated in the snippet message as well.

VXNCXNX:
Fixes #1036.

## What's broken

When `tldr` cannot be called, navi tells the user to install tealdeer:

```
navi was unable to call tldr.
Make sure tldr is correctly installed.

Note:
tldr-c-client (the default one in Homebrew) doesn't support markdown files, so navi can't use it.
The recommended client is tealdeer(https://github.com/dbrgn/tealdeer).
```

~~A user who does exactly that still gets the same message, because `client.tealdeer: true` is also required.~~

**Correction (see [the discussion below](https://github.com/denisidoro/navi/pull/1040#issuecomment-5483776960)):** that was wrong. A user who installs tealdeer gets a *different* error, and it already names the config option:

```
Failed to call:
tldr tar --markdown

Note:
The client.tealdeer config option can be set to enable tealdeer support.
```

The spawn error at `src/clients/tldr.rs:70` only fires when the binary is missing. A wrong flag means `spawn()` succeeded, so you reach the non-zero-exit branch at line 92 instead. Verified by building this branch and running both paths against a stub `tldr` that rejects `--markdown` the way tealdeer does.

So this is not a dead end, and the PR is smaller than first described: the stale repo URL, plus a `client.tealdeer` nudge in the not-installed path so the option gets set in the same step rather than after a second failure.

The linked repo has also moved:

```
$ curl -sL -o /dev/null -w "%{url_effective}\n" https://github.com/dbrgn/tealdeer
https://github.com/tealdeer-rs/tealdeer
```

## The fix

Name the config option in the disclaimer and point at the current repo. The other error path further down already mentions `client.tealdeer`, so this only brings the first one in line with it.

## Verification

```
$ PATH=/nonexistent navi --tldr ls
navi was unable to call tldr.
Make sure tldr is correctly installed.

Note:
tldr-c-client (the default one in Homebrew) doesn't support markdown files, so navi can't use it.
The recommended client is tealdeer(https://github.com/tealdeer-rs/tealdeer),
which also needs client.tealdeer: true in your config.yaml.
```

`cargo fmt --check` clean, `cargo build --release` clean, tests pass.

One note unrelated to this change: `cargo clippy --all-targets -- -D warnings` already fails on master with "this if-then-else expression returns a bool literal". I confirmed that on a clean tree.


GauravS11112003:
@denisidoro @alexis-opolka Just a followup. Thanks.

emma31-dev:
@VXNCXNX Firstly, great catch. Didn't noticed repo moved. The severity of the issue you claimed isn't there. tldr can be installed through other means. The error you face is because u don't have it in your system. See https://github.com/tealdeer-rs/tealdeer/releases to install the binary compatible with your system.

Alexis is quite busy so await a response soon

VXNCXNX:
You are right, and my PR description overstated it. I built the branch and ran both paths with a stub `tldr` that rejects `--markdown` the way tealdeer does, and they are two different messages, not one:

**tealdeer installed, `client.tealdeer` unset**

```
Failed to call:
tldr tar --markdown

Error:
error: unexpected argument '--markdown' found

Note:
The client.tealdeer config option can be set to enable tealdeer support.
```

**tldr not installed at all**

```
navi was unable to call tldr.
Make sure tldr is correctly installed.
```

The spawn error at `src/clients/tldr.rs:70` only fires when the binary is missing; a wrong flag means `spawn()` succeeded and you land on the non-zero-exit branch at line 92, which already names `client.tealdeer`. So the user who installs tealdeer does not get the same message, and does not hit a dead end. That is the part of my description that was wrong, and I have struck it through.

What is left is smaller than I wrote it up as:

- The repo URL is stale, `dbrgn/tealdeer` now redirects to `tealdeer-rs/tealdeer`. This one is unambiguous.
- The `client.tealdeer` line is a nudge in the not-installed path, so someone installing tealdeer for the first time sets the option in the same step instead of hitting the second error afterwards. In the wrong-flag path it is redundant with the note that is already there.

Happy to cut this down to the URL alone if the extra line is not worth it, that is a one-line diff. Either way, thanks for checking it properly rather than waving it through.

emma31-dev:
I'll check out the issue in `client/tldr.rs` later to confirm the issue you raised.

emma31-dev:
Wait wait wait ... There's alot of wrong assumptions being made now I read the edited description. The error u currently face isn't a source level issue but which how you installed tldr in your system. Firstly reinstall tldr from whatever Installer you use mostly then try again.

VXNCXNX:
To be clear on one thing: my tldr install isn't broken. I wrote a stub binary that rejects `--markdown` on purpose, to force that specific branch and see which message it produced. That was how I confirmed your point and withdrew my original claim.

But that's beside the point now, because none of it is in the diff. The change is two lines of a string literal: the `dbrgn/tealdeer` URL you confirmed is dead, and a mention of `client.tealdeer` in the not-installed message. No behaviour changes. If the config line is the part you'd rather not have, I'll drop it and leave just the URL.

emma31-dev:
Don't get me wrong, I have no issue with the changes but your wordings. You say you use a stub binary of tldr to force that error message that's ok but you worded it as a source level issue rather and URL change fix. You said stuff like "The spawn error at src/clients/tldr.rs:70 only fires when the binary is missing" tilting to a source level problem. Yet your fix introduces "No behaviour changes" as confirmed by you. The link https://github.com/dbrgn/tealdeer still forwards to https://github.com/tealdeer-rs/tealdeer so your change is rather optional than necessary.

PR #1043, opened 2026-08-23
fix: prevent clap from panicking on test harness arguments

phanirithvij:
Cargo supports a `--skip` argument for eg. `cargo test -- --skip test_width` but some tests can't be skipped in navi.

The issue is due to those tests evaluating the global `CONFIG` object, then clap intercepts the `--skip` argument, fails to recognize it as a CLI flag, and panics.

This pr fixes the behavior of cargo tests.


emma31-dev:
@phanirithvij  I think this can be simplified to:

if cfg!(test) {
        Self::parse_from(vec!["navi"])
 } else {
        Self::parse()
}

the e.exit() isn't good practice

emma31-dev:
close but not quite it yet. I meant:
```rust
impl ClapConfig {
    pub fn new() -> Self {
        if cfg!(test) {
            Self::parse_from(vec!["navi"])
        } else {
            Self::parse()
        }
    }
}
```

alexis-opolka:
Sorry, your PR went through my emails. It seems good to me but I can't test it yet, I might be able to this afternoon.
But I think there aren't any reason to not accept it.

